PR #15, opened 2016-06-08
update package paths

loosebazooka:


loosebazooka:
@etanshaul @meltsufin pretty trivial change, PTAL

etanshaul:
I kind of like "internal" over "hidden" but its not a big deal. The rest makes sense to me. lgtm.

loosebazooka:
ah it's hidden because of some gradle naming...

PR #16, opened 2016-06-08
Testing

loosebazooka:
Adds travis.yaml
Adds gradle testing framework support (and some tests as part of that)

https://docs.gradle.org/current/userguide/javaGradle_plugin.html
https://docs.gradle.org/current/userguide/test_kit.html#sub:test-kit-automatic-classpath-injection


loosebazooka:
@meltsufin

loosebazooka:
@patflynn @etanshaul

src/test/java/com/google/cloud/tools/gradle/appengine/BuildResultFilter.java
```diff
@@ -0,0 +1,51 @@
...
+
+import org.gradle.testkit.runner.BuildResult;
+import org.gradle.testkit.runner.BuildTask;
+
+import java.util.List;
+
+import javax.annotation.Nullable;
+
+/**
+ * Created by appu on 6/8/16.
```
etanshaul: Docs

src/test/java/com/google/cloud/tools/gradle/appengine/AppEngineStandardPluginTest.java
```diff
@@ -0,0 +1,79 @@
...
+  public void testRun_taskTree() {
+    BuildResult buildResult = GradleRunner.create()
+        .withProjectDir(testProjectDir.getRoot())
+        .withPluginClasspath()
+        .withArguments("gcpAppRun", "--dry-run")
+        .build();
+
+    final List<String> expected = Arrays.asList(":explodeWar", ":gcpAppStage", ":gcpAppRun");
+
+    Assert.assertEquals(expected, BuildResultFilter.extractTasks(buildResult, expected));
```
etanshaul: So your passing in a list of expected tasks, and then filtering them out from the actual results to see if they match. What if there are more results than expected? Is that still ok?

src/test/java/com/google/cloud/tools/gradle/appengine/AppEngineFlexiblePluginTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+//import com.google.common.io.Files;
```
etanshaul: Maybe remove this before merging?

src/test/java/com/google/cloud/tools/gradle/appengine/AppEngineStandardPluginTest.java
loosebazooka: Well yeah, when you put it like that... Anyway, it's not a great test, lemme see what I can do.

etanshaul: Yea. I'm don't know exactly what you are trying to assert but maybe instead check to see if the expected list matches exactly the tasks you get from the GradleRunner (rather than just a contains which looks like the current logic).

src/test/java/com/google/cloud/tools/gradle/appengine/AppEngineFlexiblePluginTest.java
patflynn: javadoc

src/test/java/com/google/cloud/tools/gradle/appengine/AppEngineStandardPluginTest.java
loosebazooka: My intention was to assert order and presence. But I should just check the whole list.

loosebazooka:
PTAL, added windows testing and fixed up some other stuff.

meltsufin:
Are you able to run the tests through IntelliJ? I'm getting:
`org.gradle.testkit.runner.InvalidPluginMetadataException: Test runtime classpath does not contain plugin metadata file 'plugin-under-test-metadata.properties'`

Everything works through commandline, though.

meltsufin:
The tests you've added so far mostly just verify the Gradle task trees. I think what's missing is the tests to verify the use of the common library, like the configuration of the CloudSdk and the invocations of the actions. Are you planning to add those tests later, or rely on end-to-end tests to cover that?

loosebazooka:
The tests don't work in intellij because gradle needs to run a task beforehand to set up the buildscript classpath so we can apply the plugin. see : https://docs.gradle.org/current/userguide/test_kit.html#sub:test-kit-automatic-classpath-injection

Yeah, so the test to verify the common library require some factory reconfiguration in the style that you did for maven. I figure we're going to do some design work on the common lib that will make that much more meaningful. #17 I think the end-to-end tests can cover us for now, but yeah those tests need to be added.

meltsufin:
Is it possible to run the task for setting up the buildscript classpath from the test itself? It's not absolutely necessary, but it's nice to be able to run the tests from the IDE.

loosebazooka:
I think there's a fix for this in gradle 2.14 : https://issues.gradle.org/browse/GRADLE-3433
I verified this working with gradle-2.14-rc-1, will update when they go GA : #18

In the meantime, what I've been doing is running the tests using the gradle test task instead of the intellij tests.

src/test/java/com/google/cloud/tools/gradle/appengine/AppEngineFlexiblePluginTest.java
```diff
@@ -0,0 +1,65 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ *
```
meltsufin: Document or remove comment block

meltsufin:
:ship:

PR #20, opened 2016-06-16
Integration tests

loosebazooka:
Adds integrations tests, doesn't enable them by default.

run tests using ./gradlew integTest

If you see errors with dev appserver tests, make sure it's running on Java7 (dev_appserver.py will launch using JAVA_HOME, so set that when running from the IDE).


loosebazooka:
@meltsufin

src/integTest/java/com/google/cloud/tools/gradle/appengine/AppEngineFlexiblePluginIntegTest.java
```diff
@@ -0,0 +1,74 @@
...
+import org.junit.rules.Timeout;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.Arrays;
+
+/**
+ * End to end tests for flexible environment projects
+ */
+public class AppEngineFlexiblePluginIntegTest {
```
meltsufin: I think the convention, at least for Maven integration tests, is `IT` suffix.  So, this would be `AppEngineFlexiblePluginIT`.

loosebazooka: Looking through the gradle source, I found they use IntegTest, but that might just be their style. Lemme look around some more.

loosebazooka: The google style guide seems to point towards *IntegrationTest https://google.github.io/styleguide/javaguide.html#s5.2.2-class-names

src/main/java/com/google/cloud/tools/gradle/appengine/task/io/GradleLoggerOutputListener.java
```diff
@@ -0,0 +1,42 @@
...
+
+package com.google.cloud.tools.gradle.appengine.task.io;
+
+import com.google.cloud.tools.appengine.cloudsdk.process.ProcessOutputLineListener;
+
+import org.gradle.api.logging.LogLevel;
+import org.gradle.api.logging.Logger;
+
+/**
+ * Created by appu on 6/15/16.
```
meltsufin: comment

loosebazooka: There's always one of those in my stupid commits... :disappointed:

meltsufin: maybe we should just mandate that everyone puts their name on the files they create!

src/integTest/java/com/google/cloud/tools/gradle/appengine/AppEngineFlexiblePluginIntegTest.java
meltsufin: I guess we should switch to the Google Style convention.

meltsufin:
You have subdirectory under `src` called `integTest` which looks a little odd. Why not put those tests under the normal `src/test`?

loosebazooka:
It's just the way I've seen most gradle projects setup for integrations tests (including gradle). I think partially it's because of the convention of testing, either put them somewhere else or customize the "test" task to exclude the integration tests.

meltsufin:
As long as it's a common practice for Gradle projects, it's fine.

meltsufin:
The deploy testing is timing out for me after 60 seconds.

src/main/java/com/google/cloud/tools/gradle/appengine/task/io/GradleLoggerOutputListener.java
```diff
@@ -0,0 +1,42 @@
...
+
+import com.google.cloud.tools.appengine.cloudsdk.process.ProcessOutputLineListener;
+
+import org.gradle.api.logging.LogLevel;
+import org.gradle.api.logging.Logger;
+
+/**
+ * Output listener that writes to a gradle logger
+ */
+public class GradleLoggerOutputListener implements ProcessOutputLineListener {
```
meltsufin: Should the introduction of this class and the related uses be part of a separate PR? It seems unrelated to integration testing.

loosebazooka: It's actually necessary for output capture it turns out. So there was no way to read the output unless we're redirecting to the gradle logger. The inherited redirect made all the tests fail.

meltsufin:
:ship:

PR #37, opened 2016-06-28
update pom output and include javadoc jar

loosebazooka:
most of this is just spacing @meltsufin 

the real meat : developers data in pom.xml, added javadoc-jar generation, settings.gradle to enforce project naming


meltsufin:
:ship:

PR #40, opened 2016-06-29
Only add default docker dir if it exists

loosebazooka:
@meltsufin


meltsufin:
Is this really necessary, since the plugins core lib already checks for the existence of the docker directory?

loosebazooka:
Yeah, because the way the plugin up-to-date checks work.
1. you want to mark dockerDirectory as inputDirectory so if the contents change it retriggers the stage task
2. if you mark something as an `@InputDirectory` instead of `@Input` then gradle does an existence validation... so it has to exist.

meltsufin:
Sounds good then.
:ship:

PR #41, opened 2016-06-30
Use extensions for sharing project->model

loosebazooka:
@meltsufin use gradle objects for data sharing.


meltsufin:
Is this feature somewhere in Gradle docs?

loosebazooka:
It's just what a developer at the gradle conference said was an option while they transition things.

meltsufin:
Can you just do this?
```java
project.getExtensions().add("archive", war.getArchivePath());
````
and to retrieve it...
```java
app.getStage().setArtifact((File)extensionContainer.getByName("archive"));
```

This way you don't need the `ProjectModelExtension` class.

loosebazooka:
It looks like it would technically work, but I don't think gradle expects anyone to call "add" directly because it doesn't use gradle's object instantiator to instantiate the object as a gradle object. The gradle source only seems to be using it in tests and during creation of extensions via create. Also, this way the typing is explicit and can just be extended if we need to move other data between project and model space.

meltsufin:
It's a public method. So, I would assume it's fair game to use. The only reason I would prefer it at this point is because it's more concise and makes the code easier to understand.

PR #50, opened 2016-07-18
Add java8 workaround for devAppServer

loosebazooka:


loosebazooka:
@meltsufin perhaps moving this into common library is the way to go?

meltsufin:
In the maven plugin, we only pass this flag if we detect Java 8.

loosebazooka:
makes sense, let me adjust

loosebazooka:
@meltsufin ptal

meltsufin:
:boat:

PR #62, opened 2016-09-23
v0.2.0 + remove all traces of Gradle Software Model

loosebazooka:
It appears that the Gradle Software Model for java projects is
not moving forward. Change code back to classic plugin format -
using extensions and project callbacks.

This code change affects how to configure the plugin now,
the "appengine" configuration block is now under project, not model.

OLD :
model {
  appengine {
  }
}

NEW :
appengine {
}


loosebazooka:
@patflynn @meltsufin plugin rewrite.

build.gradle
```diff
@@ -18,7 +18,7 @@ sourceCompatibility = JavaVersion.VERSION_1_7
 targetCompatibility = JavaVersion.VERSION_1_7
 
 group = 'com.google.cloud.tools'
-version = '0.1.2-SNAPSHOT'
+version = '0.2.0-SNAPSHOT'
```
patflynn: @meltsufin @loosebazooka 

Why are we still on 0.2.0 if these are 1.0 RC ready?

meltsufin: For Maven, I was thinking next release would be 1.0.0.

loosebazooka: mostly just because I change a lot of the internals and would like to play around with it for a week before switching to 1.0.0

loosebazooka: or does 1.0.0-rc basically imply the same thing?

meltsufin:
install comes up with this warning:
```
warning: [options] bootstrap class path not set in conjunction with -source 1.7
1 warning
```

loosebazooka:
Yeah cause you're probably building with java8. You can build with java7 or bypass that warning by specifying the bootstrap class in your build.gradle

```
compileJavaTask.options.bootClasspath = (new File(java7Home, "/jre/lib/rt.jar")).canonicalPath
```

meltsufin:
LGTM

PR #63, opened 2016-09-27
Add appengineShowConfiguration task

loosebazooka:
This prints out the appengine extension, the types
of all the parameters and their values after project
evaluation.

Because we can't get a description from the model anymore,
we have to write a custom task to do it.

### DIFF
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/AppEngineStandardPlugin.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/AppEngineStandardPlugin.java
@@ -20,6 +20,7 @@ package com.google.cloud.tools.gradle.appengine;
 import com.google.cloud.tools.gradle.appengine.model.AppEngineStandardExtension;
 import com.google.cloud.tools.gradle.appengine.task.CloudSdkBuilderFactory;
 import com.google.cloud.tools.gradle.appengine.task.DeployTask;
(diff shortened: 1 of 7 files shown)


loosebazooka:
@meltsufin @lesv

lesv:
Cool - I'm going to wait a bit before documenting this.

lesv:
LGTM

src/main/java/com/google/cloud/tools/gradle/appengine/AppEngineStandardPlugin.java
```diff
@@ -234,4 +238,23 @@ public void execute(Project project) {
       }
     });
   }
+
+  private void createShowConfigurationTask() {
```
meltsufin: Looks like a significant amount of code duplication here. Can this method be factored out for reuse?

lesv:
I think we have that now - https://cloud.google.com/appengine/docs/java/tools/gradle-reference  is there more you want?

src/main/java/com/google/cloud/tools/gradle/appengine/AppEngineStandardPlugin.java
loosebazooka: I mean not really, with java8 we could shorten it. But it's mostly required configuration and I prefer for it to be in the plugin configuration class.

loosebazooka:
Documentation could possibly be added using javadoc parsing, but perhaps for another day.

PR #109, opened 2017-03-21
use getProjectDir instead of getRootDir

### LINKED ISSUE #108: Plugin fails when using deep project structure
I have a project that has layout like this:

```
 - root
   | - ios
   | - android
   | - handler (in google cloud)
   | - trigger (in google cloud)
   | - etc..
```
When I run `./gradlew handler:appengineStage` from the root directory, staging fails with exception 
> Directory '[removed]/root/src/main/appengine' specified for property 'stagingConfig.appEngineDirectory' does not exist.

With plugin version 1.0 this used to work, but not anymore with 1.1. Should I have one appengine directory in the root folder and then separate appengine folders in handler and trigger folders?

All the projects are deployed to the same Google Cloud project, but represent different Services, so the cron.yaml should go to root/src/main/appengine, but the app.yaml files to project specific appengine folders? Is this correct? I have src/main/appengine folder in all projects that are destined to be deployed to Google Cloud, but not in the root project.

loosebazooka:
fixes #108

### DIFF
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/flexible/extension/StageFlexible.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/flexible/extension/StageFlexible.java
@@ -42,11 +42,11 @@ public class StageFlexible implements StageFlexibleConfiguration {
 
   public StageFlexible(Project project, File stagingDirectory) {
     this.project = project;
-    File projectRoot = project.getProjectDir();
+    File projectDir = project.getProjectDir();
 
     this.stagingDirectory = stagingDirectory;
-    this.appEngineDirectory = new File(projectRoot, "src/main/appengine");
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/cloud/tools/gradle/appengine/flexible/extension/StageFlexible.java
```diff
@@ -42,7 +42,7 @@
 
   public StageFlexible(Project project, File stagingDirectory) {
     this.project = project;
-    File projectRoot = project.getRootDir();
+    File projectRoot = project.getProjectDir();
```
akerekes: could be called `projectDir` or `projectDirectory` instead

loosebazooka: True, that might be clearer. Lemme udpate

PR #112, opened 2017-03-23
Separate yamls

loosebazooka:
allow separate appengine project configuration yaml deployments

### DIFF
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePlugin.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePlugin.java
@@ -21,6 +21,11 @@ import com.google.cloud.tools.gradle.appengine.core.extension.AppEngine;
 import com.google.cloud.tools.gradle.appengine.core.extension.Deploy;
 import com.google.cloud.tools.gradle.appengine.core.extension.Tools;
 import com.google.cloud.tools.gradle.appengine.core.task.CloudSdkBuilderFactory;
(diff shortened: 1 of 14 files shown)


src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePlugin.java
```diff
@@ -94,6 +109,96 @@ public void execute(Project project) {
     });
   }
 
+  private void createDeployCronTask() {
+    project.getTasks().create(DEPLOY_CRON_TASK_NAME, DeployCronTask.class, new Action<DeployCronTask>() {
```
akerekes: Could you use `<>`?

src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePlugin.java
```diff
@@ -94,6 +109,96 @@ public void execute(Project project) {
...
+      @Override
+      public void execute(final DeployCronTask deployTask) {
+        deployTask.setGroup(APP_ENGINE_TASK_GROUP);
+        deployTask.setDescription("Deploy Cron configuration");
+
+        project.afterEvaluate(new Action<Project>() {
```
akerekes: Could you use `<>`?

loosebazooka: I don't think it can infer type in this situation.
The signature uses a wildcard for that parameter. `<? extends Project>`

PR #121, opened 2017-04-13
Ignore gradle project when showingConfiguration

loosebazooka:


loosebazooka:
You know, I never had a test for this. Let me add that in.

akerekes:
LGTM

Issue #125, opened 2017-04-25
force-configure environment type


IuriiSergiichuk:
Hi!
As I've mentioned in #80, as there any way to force the environment type (standard/flexible) so that plugin won't complain about missing `appengine-web.xml`?

Why do I need this? - I've got copy tasks, that copy needed configurations based on the gradle argument passed, so that it would copy production/development version of `appengine-web.xml` during the build phase and right now the plugin complains about missing `appengine-web.xml` with the following exception message:
```bash
* What went wrong:
A problem occurred evaluating project ':main:webapp'.
> Could not find method run() for arguments [build_30m8mbcu9cpfcizc4mf40n9v9$_run_closure2$_closure3@47798be8] on object of type com.google.cloud.tools.gradle.appengine.core.AppEngineExtension.
```
It would not complain if I just create stub `appengine-web.xml` (even empty file with such name), but IMO it's just dirty workaround.

Could you please suggest any configuration that could be done or maybe such configuration should be implemented?

Regards,
Yuri.

P.S. My plugin config is smth like:
```gradle
appengine {
    run {
        port = 8080
        jvmFlags = ['-XX:MaxPermSize=256M', '-Xdebug', '-Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=5005']
    }
}
```
The development environment is smth like:
* Ubuntu 16.04/Windows10
* Gradle wrapper v. 3.5
* OpenJDK 1.8 / OracleJDK 1.8

loosebazooka:
I thought I had answered this before, but I guess I didn't. For now, what you want to do is apply the plugin class directly.

```
project.pluginManager.apply com.google.cloud.tools.gradle.appengine.standard.AppEngineStandardPlugin
```
instead of the `com.google.cloud.tools.appengine`

We're planning on just exposing the flex and standard plugins in a future release. But this is the best bet for now.

IuriiSergiichuk:
@loosebazooka thanks for your comment.

I'll close the issue so far.

PR #126, opened 2017-04-25
Added ability to configure Environment explicitly - #125

xSAVIKx:
There is now an option to configure environment that would be used by app-gradle plugin explicitly using following syntax:
```
appengine{
  environment = ['standard'|'flexible']
}
```

If no environment was supplied fallback to the previous behaviour would be used.
If environment doesn't match the values above configuration exception would be thrown as environment is an enum.

### DIFF
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/AppEnginePlugin.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/AppEnginePlugin.java
@@ -17,10 +17,15 @@
 
 package com.google.cloud.tools.gradle.appengine;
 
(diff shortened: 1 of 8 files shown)


loosebazooka:
Thanks for your contribution, but I don't think this would be the right way to do this. We will be allowing users to bypass auto-detection by allowing them to explicitly apply an appengine-standard or appengine-flexible plugin.

For now if you want to explicitly apply a plugin, you can do so using the class directly as described in #125

xSAVIKx:
@loosebazooka got it. I'll just close this PR.

PR #128, opened 2017-05-05
Update plugins-core to 0.3.2

loosebazooka:
This includes the ability to specify env vars for devappserver
via appengine.run.environment and for devappserver1 to read
environment from appengine-web.xml (matching how devappserver2
behaves)


src/test/java/com/google/cloud/tools/gradle/appengine/standard/AppEngineStandardExtensionTest.java
```diff
@@ -66,6 +67,18 @@ private Project setUpTestProject(String buildFileName) throws IOException {
     return p;
   }
 
+  // TODO : Make this a test that reads ALL params
```
akerekes: What does this mean?

loosebazooka: So right now it's only reading in the environment and checking that it is set correctly. Potentially we could read in every parameter and check if we're setting and getting correctly.

PR #131, opened 2017-05-09
Create a prepareRelease task

loosebazooka:
- Publish release artifacts to "${buildDir}/release" with "prepareRelease" task
- Remove gradle nexus publish option "upload"
  - code still remains in scripts/publishing.gradle

### DIFF
--- a/scripts/releasing.gradle
+++ b/scripts/releasing.gradle
@@ -1,3 +1,20 @@
+/*
+ * Copyright (c) 2017 Google Inc. All Right Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *         http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
 apply plugin: 'maven-publish'
 
 task sourceJar(type: Jar) {
@@ -51,7 +68,7 @@ task prepareRelease(type: Copy) {
     // https://discuss.gradle.org/t/gradlew-tasks-fails-in-2-4-due-to-not-finding-generatepom-task/10097/9
     tasks.realize() 
     from tasks.generatePomFileForReleasePublication
-    rename("pom-default.xml", "pom.xml")
+    rename("pom-default.xml", "pom-${project.version}.xml")
   }
 
   into "${buildDir}/release"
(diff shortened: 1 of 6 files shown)


scripts/releasing.gradle
```diff
@@ -0,0 +1,61 @@
...
+task prepareRelease(type: Copy) {
+  from jar
+  from sourceJar
+  from javadocJar
+  project.afterEvaluate {
+    // workaround for lazily generated pom publication task 
+    // https://discuss.gradle.org/t/gradlew-tasks-fails-in-2-4-due-to-not-finding-generatepom-task/10097/9
+    tasks.realize() 
+    from tasks.generatePomFileForReleasePublication
+    rename("pom-default.xml", "pom.xml")
```
akerekes: I think we need `appengine-gradle-plugin-1.3.1.pom` instead of `pom.xml`

scripts/releasing.gradle
```diff
@@ -0,0 +1,61 @@
...
+
+// this is the only way for us to create a pom using the gradle tools
+// usually this same config is used to deploy, but we're deploying
+// using another tool and really just want the pom.
+publishing {
+  publications {
+    release(MavenPublication) {
+      pom.withXml {
+        def root = asNode()
+        root.appendNode('name', 'App Engine Gradle Plugin')
```
akerekes: Is there no way to get these from an actual xml?

loosebazooka: What do you mean? It's kind of a dumb part of using the maven-publish plugin. I don't know why they did it like this. But this is how you modify the pom.

scripts/releasing.gradle
loosebazooka: okay, I can create that probably. What kind of names do you want the jar to be?

akerekes: If you look here: http://repo1.maven.org/maven2/com/google/cloud/tools/appengine-gradle-plugin/1.3.1/ the jars are properly named, only the pom needs to be follow a similar naming convention.

scripts/releasing.gradle
akerekes: I thought if we want to e.g. add a new developer, instead of changing code, just change an xml file. Maybe this is not that critical, though.

scripts/releasing.gradle
loosebazooka: Yeah, I don't know why I didn't check that first :dancer:

scripts/releasing.gradle
loosebazooka: oh, yeah perhaps, I think that's for another day.

loosebazooka:
added copyright

scripts/releasing.gradle
loosebazooka: done.

akerekes: We need the pom file to be named as `<artifactId>-<version>.pom` to conform with the repository.

loosebazooka: oh artifact id... oops

PR #132, opened 2017-05-10
Allow application of flex/standard plugins directly

loosebazooka:


patflynn:
can we add an integration test for each?

loosebazooka:
Yeah.

loosebazooka:
actually, I might need to do that as part of #133

loosebazooka:
okay, so I added necessary tests into #133

PR #133, opened 2017-05-10
Groups and tests

loosebazooka:


loosebazooka:
TestProject isn't the most perfect test project class, there's some mixing of a builder and some other stuff, but it serves the purposes of the tests pretty well. I can probably make it a more meaningful structure if we need that.

src/test/java/com/google/cloud/tools/gradle/appengine/AppEnginePluginTest.java
```diff
@@ -0,0 +1,93 @@
...
+            .addAutoDetectingBuildFile()
+            .addAppEngineWebXml()
+            .applyGradleRunner("tasks");
+
+    Assert.assertThat(
+        buildResult.getOutput(),
+        CoreMatchers.containsString(AppEngineStandardPlugin.APP_ENGINE_STANDARD_TASK_GROUP));
+    Assert.assertThat(
+        buildResult.getOutput(),
+        CoreMatchers.not(
```
akerekes: You could use static imports, and then the assertions would read more naturally: assertThat X not containsString Y.

src/test/java/com/google/cloud/tools/gradle/appengine/TestProject.java
```diff
@@ -0,0 +1,128 @@
...
+import org.gradle.api.Project;
+import org.gradle.api.internal.project.ProjectInternal;
+import org.gradle.api.plugins.JavaPlugin;
+import org.gradle.api.plugins.WarPlugin;
+import org.gradle.testfixtures.ProjectBuilder;
+import org.gradle.testkit.runner.BuildResult;
+import org.gradle.testkit.runner.GradleRunner;
+
+/** Create test projects for plugin test runs. */
+public class TestProject {
```
akerekes: If you defined this class as `org.junit.rules.ExternalResource` you could get rid of a few constructor/method calls creating an instance in test methods.

src/main/java/com/google/cloud/tools/gradle/appengine/flexible/AppEngineFlexiblePlugin.java
```diff
@@ -35,11 +35,10 @@
 /** Plugin definition for App Engine flexible environments. */
 public class AppEngineFlexiblePlugin implements Plugin<Project> {
 
+  public static final String APP_ENGINE_FLEXIBLE_TASK_GROUP = "App Engine flexible environment";
```
patflynn: flexible and standard should be capitalized.

src/test/java/com/google/cloud/tools/gradle/appengine/TestProject.java
loosebazooka: Yeah I wasn't too fond of the fact that you have to mess with TemporaryFolder when using ExternalResource with it. Maybe I'll come back to this another day.

PR #139, opened 2017-05-11
Fix path for appengine-web.xml for standard/flex detection fallback

### LINKED ISSUE #136: Fix fallback mechanism for appengine standard detection
The default location should be `src/main/webapp/WEB-INF`, not `src/main/java/webapp/WEB-INF`

https://github.com/GoogleCloudPlatform/app-gradle-plugin/blob/master/src/main/java/com/google/cloud/tools/gradle/appengine/AppEnginePlugin.java#L57

loosebazooka:
fixes #136

### DIFF
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/AppEnginePlugin.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/AppEnginePlugin.java
@@ -54,7 +54,7 @@ public class AppEnginePlugin implements Plugin<Project> {
     }
     // convention based lookup of appengine-web.xml as a fallback
     Path appengineWebXml =
-        project.getProjectDir().toPath().resolve("src/main/java/webapp/WEB-INF/appengine-web.xml");
+        project.getProjectDir().toPath().resolve("src/main/webapp/WEB-INF/appengine-web.xml");
     if (Files.exists(appengineWebXml)) {
       return true;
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/cloud/tools/gradle/appengine/AppEnginePlugin.java
```diff
@@ -54,7 +54,7 @@ private boolean isAppEngineStandard(Project project) {
     }
     // convention based lookup of appengine-web.xml as a fallback
     Path appengineWebXml =
-        project.getProjectDir().toPath().resolve("src/main/java/webapp/WEB-INF/appengine-web.xml");
+        project.getProjectDir().toPath().resolve("src/main/webapp/WEB-INF/appengine-web.xml");
```
patflynn: Can we use the war plugin to get the webapp directory?

loosebazooka: Yeah we do that, see [line 48](https://github.com/GoogleCloudPlatform/app-gradle-plugin/pull/139/files/5e7ebc6d2d2b781aaa877404222f0f5c4865310a#diff-e63a80ab373530959c0f994705e2a85cR48), this just covers the case that the war plugin is applied *after* the appengine plugin.

PR #156, opened 2017-08-08
Log Dev App Server output to file

loosebazooka:
This is useful in particular for Android Studio users who want to use the new plugin. Currently once they start the app server using "appengineStart", they don't see any output past "Dev App Server is now running". After this change, they can `tail` the output file and monitor dev-appserver output.

### DIFF
--- a/src/integTest/java/com/google/cloud/tools/gradle/appengine/AppEngineStandardPluginIntegrationTest.java
+++ b/src/integTest/java/com/google/cloud/tools/gradle/appengine/AppEngineStandardPluginIntegrationTest.java
@@ -95,7 +95,8 @@ public class AppEngineStandardPluginIntegrationTest {
       String devAppServerOutput =
           FileUtils.readFileToString(new File(expectedLogFileDir, devAppserverLogFiles[0]));
       System.out.println(devAppServerOutput);
-      Assert.assertTrue(devAppServerOutput.contains("Dev App Server is now running")
+      Assert.assertTrue(
+          devAppServerOutput.contains("Dev App Server is now running")
               || devAppServerOutput.contains("INFO:oejs.Server:main: Started"));
 
       AssertConnection.assertResponse(
--- a/src/integTest/java/com/google/cloud/tools/gradle/appengine/AppEngineStandardPluginIntegrationTest.java
+++ b/src/integTest/java/com/google/cloud/tools/gradle/appengine/AppEngineStandardPluginIntegrationTest.java
@@ -25,7 +25,6 @@ import java.io.File;
 import java.io.IOException;
 import java.util.Arrays;
 import org.apache.commons.io.FileUtils;
-import org.apache.tools.ant.DirectoryScanner;
 import org.gradle.testkit.runner.BuildResult;
 import org.gradle.testkit.runner.GradleRunner;
 import org.hamcrest.CoreMatchers;
@@ -38,17 +37,19 @@ import org.junit.rules.TemporaryFolder;
 import org.junit.rules.Timeout;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameter;
+import org.junit.runners.Parameterized.Parameters;
 
 /** End to end tests for standard projects. */
 @RunWith(Parameterized.class)
 public class AppEngineStandardPluginIntegrationTest {
 
   /** Parameterize the project source for the test. */
-  @Parameterized.Parameters
+  @Parameters
   public static Object[] data() {
-    return new Object[] {
-      "src/integTest/resources/projects/standard-project",
-      "src/integTest/resources/projects/standard-project-java8"
+    return new Object[][] {
+      {"src/integTest/resources/projects/standard-project", "Dev App Server is now running"},
+      {"src/integTest/resources/projects/standard-project-java8", "INFO:oejs.Server:main: Started"}
     };
   }
 
@@ -56,11 +57,11 @@ public class AppEngineStandardPluginIntegrationTest {
 
   @Rule public TemporaryFolder testProjectDir = new TemporaryFolder();
 
-  private final String testProjectSrcDirectory;
+  @Parameter(0)
+  public String testProjectSrcDirectory;
 
-  public AppEngineStandardPluginIntegrationTest(String testProjectSrcDirectory) {
-    this.testProjectSrcDirectory = testProjectSrcDirectory;
-  }
+  @Parameter(1)
+  public String devAppServerStartedString;
 
   @Before
   public void setUp() throws IOException {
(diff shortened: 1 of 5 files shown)


src/integTest/java/com/google/cloud/tools/gradle/appengine/AppEngineStandardPluginIntegrationTest.java
```diff
@@ -77,25 +78,41 @@ public void testDevAppServer_sync() {
...
+      GradleRunner.create()
+          .withProjectDir(testProjectDir.getRoot())
+          .withPluginClasspath()
+          .withArguments("appengineStop")
+          .build();
+
+      // give the server a couple seconds to come down
+      Thread.sleep(8000);
+
+      AssertConnection.assertUnreachable("http://localhost:8080");
```
loosebazooka: Actually I wonder if this should go outside the finally.

nkibler: It wouldn't be executed if the test fails (or something throws) if it's placed outside the finally block, so effectively it's the same as being here.

src/main/java/com/google/cloud/tools/gradle/appengine/standard/DevAppServerStartTask.java
```diff
@@ -43,13 +45,22 @@ public void setCloudSdkBuilderFactory(CloudSdkBuilderFactory cloudSdkBuilderFact
   @TaskAction
   public void startAction() throws AppEngineException, IOException {
 
+    // Add a listener to write to a file for non-blocking starts, this really only works
+    // when the gradle daemon is running (which is default for newer versions of gradle)
+    File logFile = File.createTempFile("dev_appserver", ".out", getTemporaryDir());
```
nkibler: Is it intentional that this log file could be erased after the execution of the task? The docs for `getTemporaryDir()` mention: "There are no guarantees that the contents of this directory will be kept beyond the execution of the task."

I would imagine many people may execute the task, watch it fail, and check the log to see why it failed but this file might be erased.

src/integTest/java/com/google/cloud/tools/gradle/appengine/AppEngineStandardPluginIntegrationTest.java
```diff
@@ -77,25 +78,47 @@ public void testDevAppServer_sync() {
...
+      ds.setIncludes(new String[] {"dev_appserver*.out"});
+      ds.setBasedir(expectedLogFileDir);
+      ds.scan();
+      String[] devAppserverLogFiles = ds.getIncludedFiles();
+      Assert.assertEquals(1, devAppserverLogFiles.length);
+      String devAppServerOutput =
+          FileUtils.readFileToString(new File(expectedLogFileDir, devAppserverLogFiles[0]));
+      System.out.println(devAppServerOutput);
+      Assert.assertTrue(devAppServerOutput.contains("Dev App Server is now running")
+              || devAppServerOutput.contains("INFO:oejs.Server:main: Started"));
```
nkibler: Why could this be different? It seems strange to assert something was logged _or_ something else was logged.

src/main/java/com/google/cloud/tools/gradle/appengine/standard/DevAppServerStartTask.java
loosebazooka: The assertion is to check only one log file is created. I guess I could add another test to make sure 2 log files are created for the same project. Projects start in a pristine state in this test (no build dir)

nkibler: I'm more concerned about Gradle's lifecycle management of this temporary directory, created by `getTemporaryDir()`. If there are no guarantees to it sticking around after execution of the task, would users expect it to be gone after the task is finished? Or would they expect it stays around, at least until the next task execution?

loosebazooka: oh I see what you're saying, that's a good point. It's more for temporary files. I should put this in a real place. brb with this PR

loosebazooka:
ptal

src/integTest/java/com/google/cloud/tools/gradle/appengine/AppEngineStandardPluginIntegrationTest.java
```diff
@@ -112,13 +118,6 @@ public void testDevAppServer_async() throws InterruptedException, IOException {
       Thread.sleep(8000);
```
nkibler: 8 seconds seems like a long time to wait and adding that to the test's time could be wasteful. What about a utility function that periodically checks to see if the server is down yet, and if not, blocks until it is or a specified timeout has expired?

src/main/java/com/google/cloud/tools/gradle/appengine/standard/DevAppServerStartTask.java
```diff
@@ -41,13 +43,23 @@ public void setCloudSdkBuilderFactory(CloudSdkBuilderFactory cloudSdkBuilderFact
...
+
   /** Task entrypoint : start the dev appserver (non-blocking). */
   @TaskAction
   public void startAction() throws AppEngineException, IOException {
 
     // Add a listener to write to a file for non-blocking starts, this really only works
     // when the gradle daemon is running (which is default for newer versions of gradle)
-    File logFile = File.createTempFile("dev_appserver", ".out", getTemporaryDir());
+    File logFile =
+        new File(devAppServerLoggingDir, "dev_appserver" + System.currentTimeMillis() + ".out");
```
nkibler: This will continue to build up log files until the user deletes the logging directory. This could be costly in storage without them ever realizing it; what about rewriting the same file? Or better yet, append to the same file with a max line limit? The latter may be over-engineering for this now so I think the former is the best option.

loosebazooka: So I think it's pretty common that gradle/maven users run clean somewhat often (habit, probably), but I was thinking about this over the weekend, specifically users who wish to have a post-run task in an IDE can trigger a log reader to launch if the log file name is consistent. 
The downside is that old logs will be erased, errors should be repeatable though (and gradle users can presumably trigger a preserve logs task on their own before running a new appserver).

So yeah, lets overwrite the file.

src/integTest/java/com/google/cloud/tools/gradle/appengine/AppEngineStandardPluginIntegrationTest.java
loosebazooka: Yeah this test is pretty old, not sure why we picked this, probably because at the time it was a lot of work (when is it not?). Anyway, I can try to fix this.

src/integTest/java/com/google/cloud/tools/gradle/appengine/AppEngineStandardPluginIntegrationTest.java
loosebazooka: I took out this finally, it's not helpful when the first part of the test fails and then this fails, since the first exception is lost.

PR #162, opened 2017-09-04
Use sync instead of copy on the explodeWar task

int128:
This pull request improves processing time and battery consumption for the `explodeWar` task by using [sync task](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.Sync.html). Current implementation always copies all files (big JARs!) but this copies only modified files.

### DIFF
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/standard/ExplodeWarTask.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/standard/ExplodeWarTask.java
@@ -18,49 +18,15 @@
 package com.google.cloud.tools.gradle.appengine.standard;
 
 import java.io.File;
-import org.gradle.api.Action;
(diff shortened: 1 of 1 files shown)


loosebazooka:
Interesting idea, we do like to clean the exploded app directory though. Does sync remove files in the destination that are NOT in the source?

int128:
Yes, sync task removes files not in the source. For example, `build/exploded-app/WEB-INF/classes/foo` will be removed unless `src/main/resources/foo` exists.

>This task is like the Copy task, except the destination directory will only contain the files copied. All files that exist in the destination directory will be deleted before copying files, unless a Sync.preserve(org.gradle.api.Action) is specified.
>
>https://docs.gradle.org/current/dsl/org.gradle.api.tasks.Sync.html

loosebazooka:
Thanks @int128 for helping improve the codebase. I'll merge this in.

PR #164, opened 2017-09-13
preserve datastore-indexes-auto across non-clean rebuilds

### LINKED ISSUE #165: Preserve datastore-indexes-auto.xml across non-clean rebuilds
Since we're using the Sync task now, make sure we preserve the datastore-indexes-auto.xml file which generates indexes based on dev-appserver runs.

see : https://cloud.google.com/appengine/docs/standard/java/config/indexconfig for more info on datastore-indexes-auto.xml

loosebazooka:
fixes #165


coollog:
Is there an Issue that goes along with this?

loosebazooka:
@coollog updated description.

PR #170, opened 2017-09-26
Checks the Gradle version to make sure it is up to date.

### LINKED ISSUE #169: Add minimum gradle version check
Something like this : https://github.com/GoogleCloudPlatform/gradle-appengine-plugin/blob/master/src/main/groovy/com/google/appengine/AppEnginePlugin.groovy#L176

since we use features in newer versions of gradle (like Sync)

Issues popping up : https://groups.google.com/forum/#!topic/google-appengine/TSglEs0voO8

coollog:
Fixes #169 .

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
+## 1.3.4
+
+### Added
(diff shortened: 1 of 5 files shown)


coollog:
Updated the CHANGELOG with this change.

PR #176, opened 2017-10-20
Upgrade appengine-plugins-core to 0.3.9

chanseokoh:
We need to pick up fixes in recent releases.


chanseokoh:
Okay, simple upgrading doesn't work.

chanseokoh:
@loosebazooka can you take on this later? It's not urgent.

PR #183, opened 2017-10-25
Adds a prepare_release.sh script.

### LINKED ISSUE #178: update release process to maven central.
Looks like we're going to hold off releases to gradle plugin portal until they fix their download tracking.
So lets automate this release the same way as the maven plugin release.

coollog:
Fixes #178 . Similar to https://github.com/GoogleCloudPlatform/appengine-plugins-core/issues/479.


loosebazooka:
do you have to make this executable too ?

loosebazooka:
actually this should have the gradle release plugin already.

coollog:
Yea, this one I have the permissions set correctly.

It uses the gradle release plugin. I wrap it with this script so that the release instructions for both the Maven and Gradle plugins are the same.

scripts/prepare_release.sh
```diff
@@ -0,0 +1,53 @@
...
+fi
+
+# Checks out a new branch for this version release (eg. 1.5.7).
+git checkout -b release_v${VERSION}
+
+# Changes the version for release and creates the commits/tags.
+echo | ./gradlew release
+
+# Pushes the release branch to Github.
+git push --tags --set-upstream origin release_v${VERSION}
```
loosebazooka: ./gradlew release pushes the tag automatically.

loosebazooka: I would rather not push other tags up with it.

coollog: Okay, removed the --tags.

PR #203, opened 2018-02-28
Begin implementing Cloud SDK installer (incomplete)

TadCordle:
Adds skeleton for a "managedSdkTask". So far it just prints out the configuration parameters it needs.


src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
```diff
@@ -79,6 +82,32 @@ public void execute(Project project) {
...
+                        new Action<Project>() {
+                          @Override
+                          public void execute(Project p) {
+                            managedSdkTask.setToolsConfig(toolsExtension);
+                          }
+                        });
+                  }
+                });
+
+    project.getTasks().getByName(BasePlugin.ASSEMBLE_TASK_NAME).dependsOn(managedSdkTask);
```
coollog: I'm probably just missing something, but what is the purpose of attaching the assemble task to the managedSdkTask?

src/main/java/com/google/cloud/tools/gradle/appengine/core/ManagedSdkTask.java
```diff
@@ -0,0 +1,53 @@
+package com.google.cloud.tools.gradle.appengine.core;
```
coollog: In IntelliJ, in `Preferences -> Editor -> File and Code Templates`, for the `Class`, `Interface`, and `Enum` files, you could insert the copyright notice to the top of the template so that it is automatically added to each new file.

src/main/java/com/google/cloud/tools/gradle/appengine/core/ManagedSdkTask.java
```diff
@@ -0,0 +1,53 @@
...
+    getProject()
+        .getLogger()
+        .lifecycle("Download Cloud SDK: " + toolsExtension.getDownloadCloudSdk());
+    getProject().getLogger().lifecycle("Cloud SDK Home: " + toolsExtension.getCloudSdkHome());
+
+    String sdkVersion = toolsExtension.getCloudSdkVersion();
+    File sdkHome = toolsExtension.getCloudSdkHome();
+    boolean download = toolsExtension.getDownloadCloudSdk();
+
+    if (download) {
```
coollog: A good way to write this skeleton may be to write an interface that is used by `ManagedSdkTask` in these `TODOs` so that the full code here is written out, but the implementation of the interface is not.

For example, we could have a class like:

```java
class Orchestra {
  private final Conductor conductor;
  void perform() {
    conductor.prepare();
    conductor.conduct();
    conductor.stop();
  }
}
```

Where `Conductor` is just an interface.
When `Conductor` is implemented, it could either be implemented with a separate implementation class, or just be converted to a class itself.

The benefit of this approach is that:

1) There is no unfilled skeleton code that might be accidentally used. (Having an unimplemented interface ensures that the code would not run.)
2) The code is testable - the interface can be mocked (with like Mockito) to run test-specific code so that the logic can be tested. Later, when the interface is implemented, you'll only have to write a test for that interface.

src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
TadCordle: This was a bit of a guess on my part; I wanted to make sure the cloud sdk is downloaded before the project as built, so I have it run before the "assemble" task. Is there a better place to put it in the task tree?

coollog: Hmm, I believe there isn't a need to have assemble happen after the managedCloudSdk task (building the project artifacts shouldn't depend on there being Cloud SDK). However, the other tasks in this plugin would need Cloud SDK to be present. @loosebazooka should there be a hard dependency from the other tasks to the managedCloudSdk task?

loosebazooka: Yeah it might not actually be ideal. We could probably have the tasks we know that need the cloud sdk just depend on this task.

like deploy/run/stage, there's also a way to write task addition callbacks. so when any task that is added that starts with "appengine", you make it depend on managedCloudSdk.

src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
```diff
@@ -79,6 +82,32 @@ public void execute(Project project) {
         });
   }
 
+  private void createManagedSdkTask() {
+    ManagedSdkTask managedSdkTask =
```
loosebazooka: I realize I wrote mangedSdkTask in the doc, but perhaps this should be downloadCloudSdk or something.

src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
```diff
@@ -89,7 +89,8 @@ private void createManagedSdkTask() {
 
               project.afterEvaluate(
                   p -> {
-                    downloadCloudSdkTask.setToolsConfig(toolsExtension);
+                    downloadCloudSdkTask.setToolsExtension(toolsExtension);
+                    downloadCloudSdkTask.setCloudSdkBuilderFactor(cloudSdkBuilderFactory);
```
coollog: looks like a typo on `Factor`

src/main/java/com/google/cloud/tools/gradle/appengine/core/SdkDownloader.java
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.google.cloud.tools.gradle.appengine.core;
+
+import java.io.File;
+
+public interface SdkDownloader {
```
coollog: I think this can be package-private?

src/main/java/com/google/cloud/tools/gradle/appengine/core/SdkDownloader.java
```diff
@@ -0,0 +1,27 @@
...
+
+  File downloadSdk(String version);
+
+  boolean sdkIsValid(String version, File home);
```
coollog: I believe a common convention we use is to name booleans with `is` at the start - like `isSdkValid`

src/main/java/com/google/cloud/tools/gradle/appengine/core/SdkDownloader.java
coollog: nit: might be good to call use `CloudSdk` since that is the name used in other parts of the package.

src/main/java/com/google/cloud/tools/gradle/appengine/core/SdkDownloader.java
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright (c) 2016 Google Inc. All Right Reserved.
```
coollog: s/2016/2018/

src/test/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTaskTest.java
```diff
@@ -0,0 +1,159 @@
+/*
+ * Copyright (c) 2016 Google Inc. All Right Reserved.
```
coollog: s/2016/2018/

src/test/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTaskTest.java
```diff
@@ -0,0 +1,159 @@
...
+    exception.expect(GradleException.class);
+    exception.expectMessage(
+        "Specified Cloud SDK version and actual version of the SDK installed in the "
+            + "specified directory do not match. You must either specify the correct "
+            + "cloudSdkHome and cloudSdkVersion, or you can remove the cloudSdkHome field "
+            + "to download the version you want.");
+
+    downloadCloudSdkTask.downloadCloudSdkAction();
+    verify(sdkDownloader, never()).downloadSdk(version);
+    verify(sdkDownloader, times(1)).sdkIsValid(version, home);
```
coollog: I believe `times(1)` is the default.

src/test/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTaskTest.java
```diff
@@ -0,0 +1,159 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+import org.junit.rules.TemporaryFolder;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.junit.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class DownloadCloudSdkTaskTest {
```
coollog: Usually, we put a class javadoc on a test like this: `/** Tests for {@link DownloadCloudSdkTask}. */`

src/test/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTaskTest.java
```diff
@@ -0,0 +1,159 @@
...
+
+  @Mock private ToolsExtension toolsExtension;
+  @Mock private CloudSdkBuilderFactory cloudSdkBuilderFactory;
+  @Mock private SdkDownloader sdkDownloader;
+
+  @Rule public ExpectedException exception = ExpectedException.none();
```
coollog: Oh this is neat!

src/test/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTaskTest.java
```diff
@@ -0,0 +1,159 @@
...
+    File home = getTempHomeDirectory("LATEST");
+    when(toolsExtension.getCloudSdkVersion()).thenReturn(version);
+    when(toolsExtension.getCloudSdkHome()).thenReturn(home);
+
+    // Assume SDK installation is valid
+    when(sdkDownloader.sdkIsValid(version, home)).thenReturn(true);
+
+    downloadCloudSdkTask.downloadCloudSdkAction();
+    verify(sdkDownloader, never()).downloadSdk(version);
+    verify(sdkDownloader, times(1)).sdkIsValid(version, home);
```
coollog: nit: Generally, only methods that have side effects are necessary to verify their calls.

src/test/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTaskTest.java
```diff
@@ -0,0 +1,159 @@
...
+    verify(cloudSdkBuilderFactory, times(1)).setCloudSdkHome(home);
```
coollog: Might be good in these tests to assert the CloudSdkHome set on `cloudSdkBuilderFactory` rather than verifying a method call.

src/test/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTaskTest.java
TadCordle: Left this out for now, it was causing a conflict between checkstyle and verifyGoogleJavaFormat for some reason.

src/main/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTask.java
```diff
@@ -33,7 +33,7 @@ public void setToolsExtension(ToolsExtension toolsExtension) {
     this.toolsExtension = toolsExtension;
   }
 
-  public void setCloudSdkBuilderFactor(CloudSdkBuilderFactory cloudSdkBuilderFactor) {
+  public void setCloudSdkBuilderFactory(CloudSdkBuilderFactory cloudSdkBuilderFactor) {
```
coollog: Looks like the parameter name is still `cloudSdkBuilderFactor`.

src/main/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTask.java
```diff
@@ -66,6 +66,6 @@ public void downloadCloudSdkAction() {
       }
     }
 
-    cloudSdkBuilderFactory.setCloudSdkHome(sdkHome);
+    this.cloudSdkBuilderFactory.setCloudSdkHome(sdkHome);
```
coollog: I believe `this` is not necessary if `cloudSdkBuilderFactory` is not also a local variable.

src/test/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTaskTest.java
TadCordle: I talked to Appu about this and left it as a verify() for now for a few reasons.
1. setCloudSdkHome() is simple enough that verifying the method is called and asserting that the value was actually set are equally useful.
2. The fact that I'm using a mock CloudSdkBuilderFactory meant the factory I was trying to update wasn't the real one, which makes verifying it this way a bit tricky.
3. Actually checking the factory's home value requires adding another getter.
4. These tests are mostly for checking the branching logic of DownloadCloudSdkTask, so I don't think it's within the scope of these tests.

src/test/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTaskTest.java
TadCordle: Should I change this in every file in the project, or just the ones I've edited since I started? There are quite a few that still say 2016.

coollog: The copyright date indicates the year that the file was written. Essentially, if someone were to copy the code, they would only violate the copyright if that was done after the copyright year. Therefore, any existing files should keep the original copyright date they were created in.

PR #208, opened 2018-03-05
Implement CloudSdkDownloader

### LINKED ISSUE #205: Implement CloudSdkDownloader
The `downloadCloudSdk` task doesn't actually do anything yet.

TadCordle:
Fixes #205

### DIFF
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/core/CloudSdkDownloader.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/core/CloudSdkDownloader.java
@@ -17,11 +17,199 @@
 
 package com.google.cloud.tools.gradle.appengine.core;
 
+import com.google.cloud.tools.appengine.cloudsdk.CloudSdk;
+import com.google.cloud.tools.appengine.cloudsdk.CloudSdkNotFoundException;
+import com.google.cloud.tools.appengine.cloudsdk.CloudSdkOutOfDateException;
+import com.google.cloud.tools.appengine.cloudsdk.CloudSdkVersionFileException;
+import com.google.cloud.tools.managedcloudsdk.BadCloudSdkVersionException;
+import com.google.cloud.tools.managedcloudsdk.ConsoleListener;
+import com.google.cloud.tools.managedcloudsdk.ManagedCloudSdk;
+import com.google.cloud.tools.managedcloudsdk.ManagedSdkVerificationException;
+import com.google.cloud.tools.managedcloudsdk.ManagedSdkVersionMismatchException;
+import com.google.cloud.tools.managedcloudsdk.ProgressListener;
+import com.google.cloud.tools.managedcloudsdk.UnsupportedOsException;
+import com.google.cloud.tools.managedcloudsdk.Version;
+import com.google.cloud.tools.managedcloudsdk.command.CommandExecutionException;
+import com.google.cloud.tools.managedcloudsdk.command.CommandExitException;
+import com.google.cloud.tools.managedcloudsdk.components.SdkComponent;
+import com.google.cloud.tools.managedcloudsdk.components.SdkComponentInstaller;
+import com.google.cloud.tools.managedcloudsdk.install.SdkInstaller;
+import com.google.cloud.tools.managedcloudsdk.install.SdkInstallerException;
+import com.google.cloud.tools.managedcloudsdk.update.SdkUpdater;
 import java.io.File;
+import java.io.IOException;
+import org.gradle.api.GradleException;
+import org.gradle.api.logging.Logger;
 
-interface CloudSdkDownloader {
+public class CloudSdkDownloader {
 
-  File downloadSdk(String version);
+  private Logger logger;
 
-  boolean isSdkValid(String version, File home);
+  private SdkInstaller installer;
+  private SdkComponentInstaller componentInstaller;
+  private SdkUpdater updater;
+
+  public CloudSdkDownloader() {
+    this.logger = null;
+  }
+
+  public CloudSdkDownloader(Logger logger) {
+    this.logger = logger;
+  }
+
+  public void setInstaller(SdkInstaller installer) {
+    this.installer = installer;
+  }
+
+  public void setComponentInstaller(SdkComponentInstaller componentInstaller) {
+    this.componentInstaller = componentInstaller;
+  }
+
+  public void setUpdater(SdkUpdater updater) {
+    this.updater = updater;
+  }
+
+  /**
+   * Download the SDK at specified version, or update existing version.
+   *
+   * @return the downloaded SDK's home directory.
+   * @throws GradleException when installation fails.
+   */
+  public File downloadSdk(String version) {
+    // Create ManagedCloudSdk from given version
+    ManagedCloudSdk sdk;
+    try {
+      if (version.equals(Version.LATEST.getVersion())) {
+        sdk = ManagedCloudSdk.newManagedSdk();
+      } else {
+        sdk = ManagedCloudSdk.newManagedSdk(new Version(version));
+      }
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/cloud/tools/gradle/appengine/core/CloudSdkDownloader.java
```diff
@@ -17,11 +17,199 @@
...
+    ConsoleListener consoleListener =
+        new ConsoleListener() {
+          @Override
+          public void console(String rawString) {}
+        };
+
+    try {
+      // Install sdk if not installed
+      if (!sdk.isInstalled()) {
+        if (installer == null) {
```
coollog: This this `null`-check could potentially cause a bug if you re-use the `downloadSdk` method, but the actual `ManagedCloudSdk` used is different (should have created a new `Installer` but used an old one).

This `null`-check seems to be used for just testing purposes, where you use a mocked `Installer` that gets set as the `installer` field. It's generally a good idea not to have logic in your code that is only for the purpose of testing (in other words, generally, [tests should be blind to the actual implementation of the method they are testing](https://testing.googleblog.com/2013/08/testing-on-toilet-test-behavior-not.html) - method call verifications violate this and should really be avoided unless necessary).

A possible solution here might be to mock the `ManagedCloudSdk#newInstaller` and return your mocked `Installer`.

src/main/java/com/google/cloud/tools/gradle/appengine/core/CloudSdkDownloader.java
```diff
@@ -17,11 +17,199 @@
...
+  }
+
+  /**
+   * Validates the Cloud SDK.
+   *
+   * @return true if the sdk in specified home directory is valid.
+   */
+  public boolean isSdkValid(String version, File home) {
+    CloudSdk cloudSdk = new CloudSdk.Builder().sdkPath(home.toPath()).build();
+    if (!version.equals("LATEST") && !version.equals(cloudSdk.getVersion().toString())) {
```
coollog: Maybe use `Version.LATEST.getVersion()` like on line 164?

src/main/java/com/google/cloud/tools/gradle/appengine/core/CloudSdkDownloader.java
```diff
@@ -17,11 +17,199 @@
...
+      if (logger != null) {
+        logger.warn(
+            "Cloud SDK validate: Specified version ("
+                + version.toString()
```
coollog: nit: `toString` is not necessary when using the string + operator.

src/test/java/com/google/cloud/tools/gradle/appengine/core/CloudSdkDownloaderTest.java
```diff
@@ -0,0 +1,133 @@
...
+      downloader.setInstaller(getTempInstaller(Version.LATEST));
+      downloader.setComponentInstaller(getTempComponentInstaller());
+      downloader.setUpdater(getTempUpdater());
+
+      File sdkHome = downloader.downloadSdk(Version.LATEST.toString());
+
+      Assert.assertTrue(sdkHome.exists());
+      Assert.assertTrue(getTempGcloud().toFile().exists());
+    } catch (IOException | UnsupportedOsException e) {
+      Assert.fail();
```
coollog: Exceptions that should fail the test can just be left uncaught and added to the throws list of the method declaration.

src/test/java/com/google/cloud/tools/gradle/appengine/core/CloudSdkDownloaderTest.java
```diff
@@ -0,0 +1,133 @@
...
+import org.gradle.internal.impldep.org.junit.Rule;
+import org.gradle.internal.impldep.org.junit.Test;
+import org.junit.Assert;
+import org.junit.rules.TemporaryFolder;
+
+public class CloudSdkDownloaderTest {
+
+  @Rule TemporaryFolder temporaryFolder = new TemporaryFolder();
+
+  private CloudSdkDownloader downloader;
```
coollog: It looks like this field can just be a local variable, or just set to `new CloudSdkDownloader()`.

src/test/java/com/google/cloud/tools/gradle/appengine/core/CloudSdkDownloaderTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+  private SdkComponentInstaller getTempComponentInstaller() throws UnsupportedOsException {
+    return SdkComponentInstaller.newComponentInstaller(getTempGcloud());
+  }
+
+  private SdkUpdater getTempUpdater() throws UnsupportedOsException {
+    return SdkUpdater.newUpdater(getTempGcloud());
+  }
+
+  private Path getTempGcloud() throws UnsupportedOsException {
```
coollog: It looks like this method can be replaced with `ManagedCloudSdk#getGcloud`? (Relativized from `ManagedCloudSdk#getSdkHome`)

src/test/java/com/google/cloud/tools/gradle/appengine/core/CloudSdkDownloaderTest.java
```diff
@@ -0,0 +1,133 @@
...
+    OsInfo info = OsInfo.getSystemOsInfo();
+    String gcloudBin = info.name().equals(Name.WINDOWS) ? "gcloud.cmd" : "gcloud";
+    return temporaryFolder.getRoot().toPath().resolve("bin").resolve(gcloudBin);
```
coollog: It seems that this is resolving a `bin` folder in the temporaryFolder's root - was it supposed to use the `temporaryFolder.newFolder` created by `getTempInstaller`?

src/main/java/com/google/cloud/tools/gradle/appengine/core/CloudSdkDownloader.java
```diff
@@ -17,11 +17,199 @@
...
+import org.gradle.api.logging.Logger;
 
-interface CloudSdkDownloader {
+public class CloudSdkDownloader {
 
-  File downloadSdk(String version);
+  private Logger logger;
 
-  boolean isSdkValid(String version, File home);
+  private SdkInstaller installer;
```
loosebazooka: We should be able to mock these out without creating these member variables. If we inject ManagedCloudSdk into this class, we can just mock the return values for the methods newInstaller, newUpdater and so on to return mocks.

src/main/java/com/google/cloud/tools/gradle/appengine/core/CloudSdkDownloader.java
```diff
@@ -17,11 +17,199 @@
...
+
+    return sdk.getSdkHome().toFile();
+  }
+
+  /**
+   * Validates the Cloud SDK.
+   *
+   * @return true if the sdk in specified home directory is valid.
+   */
+  public boolean isSdkValid(String version, File home) {
```
loosebazooka: If this class is only validating SDKs that are user specified, it does not appear like we should include it as part of the sdk downloader class. Validation might even need to happen in another task (checkCloudSdk)?

src/main/java/com/google/cloud/tools/gradle/appengine/core/CloudSdkDownloader.java
loosebazooka: I just realized how many times I said mock in this comment.

Basically, find a way to pass in the ManagedCloudSdk at construction time. And in our tests, just mock the internals out.

TadCordle:
I'm going to refactor a bunch and submit a new pull request in the future; Appu and I decided it'd be best to split the installation and validation into two separate tasks, plus the way things are written right now make things really hard to test.

PR #213, opened 2018-03-07
Implement sdk downloader

TadCordle:
Some (hopefully) improvements over the last PR:
- Validation and installation are going to be handled by separate tasks.
- The actual installation functionality takes place within the `DownloadCloudSdkTask` class instead of being abstracted away in another layer.
- The tests don't actually install anything, they just verify the logic. Which a) speeds things up, and b) doesn't require dealing with the mess of passing temporary directories around.

Still needs fixing:
- Validation task is coming in another PR (this one only has the installation task).
- `ProgressListener` and `ConsoleListener` still need to be passed in instead of using anonymous classes.


src/main/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTask.java
```diff
@@ -37,35 +42,84 @@ public void setCloudSdkBuilderFactory(CloudSdkBuilderFactory cloudSdkBuilderFact
...
-          throw new GradleException(
-              "Specified Cloud SDK version and actual version of the SDK installed in the "
-                  + "specified directory do not match. You must either specify the correct "
-                  + "cloudSdkHome and cloudSdkVersion, or you can remove the cloudSdkHome field "
-                  + "to download the version you want.");
-        }
-      }
+    managedCloudSdk = managedCloudSdkFactory.newManagedSdk(sdkVersion);
+
+    ProgressListener progressListener =
```
loosebazooka: I think we probably want to pass these in to this task. (`ProgressListener` and `ConsoleListener`). File a bug to do this in a follow up PR.

src/main/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTask.java
```diff
@@ -37,35 +42,84 @@ public void setCloudSdkBuilderFactory(CloudSdkBuilderFactory cloudSdkBuilderFact
...
-  public void setSdkDownloader(CloudSdkDownloader downloader) {
-    this.downloader = downloader;
+  public void setManagedCloudSdkFactory(ManagedCloudSdkFactory managedCloudSdkFactory) {
+    this.managedCloudSdkFactory = managedCloudSdkFactory;
   }
 
   /** Task entrypoint : Download/update/verify Cloud SDK installation. */
   @TaskAction
-  public void downloadCloudSdkAction() {
+  public void downloadCloudSdkAction() throws Exception {
```
coollog: I think we tend to avoid using the superclass `Exception` and just include all the individual exceptions.

src/test/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTaskTest.java
```diff
@@ -56,96 +61,50 @@ public void setup() {
...
 
-    when(sdkDownloader.isSdkValid(version, home)).thenReturn(true);
-
-    downloadCloudSdkTask.downloadCloudSdkAction();
-    verify(sdkDownloader, never()).downloadSdk(version);
-    verify(cloudSdkBuilderFactory).setCloudSdkHome(home);
+    when(managedCloudSdk.newInstaller()).thenReturn(installer);
+    when(managedCloudSdk.newComponentInstaller()).thenReturn(componentInstaller);
+    when(managedCloudSdk.newUpdater()).thenReturn(updater);
+    when(managedCloudSdk.getSdkHome()).thenReturn(new File("").toPath());
```
coollog: `new File("").toPath()` can be shortened to just `Paths.get("")`.

src/main/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTask.java
```diff
@@ -37,35 +42,84 @@ public void setCloudSdkBuilderFactory(CloudSdkBuilderFactory cloudSdkBuilderFact
...
+    }
+
+    // If version is set to LATEST, update Cloud SDK
+    if (!managedCloudSdk.isUpToDate()) {
+      SdkUpdater updater = managedCloudSdk.newUpdater();
+      updater.update(progressListener, consoleListener);
     }
 
-    cloudSdkBuilderFactory.setCloudSdkHome(sdkHome);
+    cloudSdkBuilderFactory.setCloudSdkHome(managedCloudSdk.getSdkHome().toFile());
```
loosebazooka: I didn't realize this earlier, but we probably don't want to be modifying extension properties from within a task execution. We might need to set this outside somewhere, probably in the AppEngineCorePluginConfiguration class.

And in that case, we're probably going to need to instantiate an instance of ManageCloudSdk out there too. You can file bugs to track this change, because it's probably going to mean that we don't pass a ManagedCloudSdkFactory in here, but just a ManagedCloudSdk instead. So maybe do it in a follow up PR.

src/test/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTaskTest.java
```diff
@@ -56,96 +61,50 @@ public void setup() {
...
-    File tempDir = getTempHomeDirectory(version);
-    when(sdkDownloader.downloadSdk(version)).thenReturn(tempDir);
+    when(managedCloudSdkFactory.newManagedSdk(version)).thenReturn(managedCloudSdk);
 
+    when(managedCloudSdk.isInstalled()).thenReturn(true);
+    when(managedCloudSdk.hasComponent(SdkComponent.APP_ENGINE_JAVA)).thenReturn(false);
     downloadCloudSdkTask.downloadCloudSdkAction();
-    verify(sdkDownloader).downloadSdk(version);
-    verify(cloudSdkBuilderFactory).setCloudSdkHome(tempDir);
+    verify(managedCloudSdk, times(0)).newInstaller();
```
coollog: `times(0)` can be `never()` I believe

src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
```diff
@@ -69,11 +70,12 @@ public void configureCoreProperties(
     createShowConfigurationTask();
   }
 
-  private void configureCloudSdkBuilderFactory() {
+  private void configureFactories() {
     project.afterEvaluate(
         project -> {
           // create the sdk builder factory after we know the location of the sdk
           cloudSdkBuilderFactory = new CloudSdkBuilderFactory(toolsExtension.getCloudSdkHome());
+          managedCloudSdkFactory = new ManagedCloudSdkFactory();
```
coollog: Maybe this could parallel the `CloudSdkBuilderFactory` instantiation by getting `toolsExtension.getCloudSdkVersion()` here.

src/main/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTask.java
coollog: This should probably be in a separate class or static to avoid cluttering the `downloadCloudSdkAction` method.

src/main/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTask.java
TadCordle: There are a lot of exceptions that can be thrown (IntelliJ's autocomplete shortened the list to just "Exception"), should I still add individual exceptions even when the list is really long?

src/main/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTask.java
TadCordle: Will handle this in a different PR (probably when I implement better progress display).

src/main/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTask.java
coollog: Hmm, not sure then. @loosebazooka @elharo ?

loosebazooka: This code is launched by the gradle system, all exception get wrapped as GradleExceptions or TaskException or something I think and effectively end the build process? We'll never really be handling what's thrown here, so I believe there is loss of information. If on the other hand we did want to handle everything then we should have the appropriate try/catch blocks in the code here.

elharo: Yes, I definitely prefer explicit exception lists where possible. It's OK to wrap lower level exceptions in a supertype if the supertype is itself thrown. For instance, you don't need to declare FileNotFoundException if a raw IOException can also be thrown. However you shouldn't declare that IOException can be thrown when it isn't. java.lang.Exception should generally not be declared excpet in very special circumstances such as Callable or Runnable where anything can happen, and probably not then.

src/main/java/com/google/cloud/tools/gradle/appengine/core/ManagedCloudSdkFactory.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright (c) 2018 Google Inc. All Right Reserved.
```
elharo: we tend not to use (c) in our copyrights and arguably it's not correct legal form: https://g3doc.corp.google.com/company/teams/opensource/copyright.md?cl=head

src/test/java/com/google/cloud/tools/gradle/appengine/core/ManagedCloudSdkFactoryTest.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright (c) 2018 Google Inc. All Right Reserved.
```
elharo: No (c)

src/test/java/com/google/cloud/tools/gradle/appengine/core/ManagedCloudSdkFactoryTest.java
```diff
@@ -0,0 +1,45 @@
...
+import com.google.cloud.tools.managedcloudsdk.UnsupportedOsException;
+import com.google.cloud.tools.managedcloudsdk.Version;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class ManagedCloudSdkFactoryTest {
+
+  @Test
+  public void testNewManagedSdk() throws UnsupportedOsException, BadCloudSdkVersionException {
+    ManagedCloudSdk sdk;
```
elharo: declare and initialize on the same line

src/test/java/com/google/cloud/tools/gradle/appengine/core/ManagedCloudSdkFactoryTest.java
```diff
@@ -0,0 +1,45 @@
...
+    ManagedCloudSdkFactory factory = new ManagedCloudSdkFactory();
+
+    // There's no way of testing for direct ManagedCloudSdk equality, so compare home paths
+    sdk = factory.newManagedSdk(null);
+    Assert.assertEquals(ManagedCloudSdk.newManagedSdk().getSdkHome(), sdk.getSdkHome());
+
+    sdk = factory.newManagedSdk("LATEST");
```
elharo: These should be three separate tests unless you're deliberately testing something about the order of these calls.

src/main/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTask.java
TadCordle: Filed bug #216

src/main/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTask.java
```diff
@@ -17,55 +17,114 @@
...
-                  + "specified directory do not match. You must either specify the correct "
-                  + "cloudSdkHome and cloudSdkVersion, or you can remove the cloudSdkHome field "
-                  + "to download the version you want.");
-        }
-      }
+  public void downloadCloudSdkAction()
+      throws UnsupportedOsException, BadCloudSdkVersionException, ManagedSdkVerificationException,
+          ManagedSdkVersionMismatchException, InterruptedException, CommandExecutionException,
+          SdkInstallerException, CommandExitException, IOException {
+    managedCloudSdk = managedCloudSdkFactory.newManagedSdk();
```
coollog: This looks like it could be a local variable?

src/main/java/com/google/cloud/tools/gradle/appengine/core/ManagedCloudSdkFactory.java
loosebazooka: I realize I told you to do this wholesale on the project, but I think protocol is to fix this in this PR and then fix the non-PR related files in a followup.

src/main/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTask.java
TadCordle: Fixed

loosebazooka:
LGTM, you can probably submit this now.

coollog:
Oh, I've been using `Inc.` everywhere in Jib. Looks like I'll be changing to LLC. from now on too.

PR #218, opened 2018-03-08
Implement checkCloudSdk task

### LINKED ISSUE #212: Split SDK installation and SDK validation into two separate tasks
Either a `downloadCloudSdk` or `checkCloudSdk` task should run depending on the tools extension parameters that are specified in build.gradle, rather than having each handled by the same task.

TadCordle:
Fixes #212

### DIFF
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/core/CheckCloudSdkTask.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/core/CheckCloudSdkTask.java
@@ -41,7 +41,7 @@ public class CheckCloudSdkTask extends DefaultTask {
     this.cloudSdkBuilderFactory = cloudSdkBuilderFactory;
   }
 
-  /** Task entrypoint : Download/update/verify Cloud SDK installation. */
+  /** Task entrypoint : Verify Cloud SDK installation. */
   @TaskAction
   public void checkCloudSdkAction() {
     File home = toolsExtension.getCloudSdkHome();
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/core/CheckCloudSdkTask.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/core/CheckCloudSdkTask.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018 Google Inc. All Right Reserved.
+ * Copyright 2018 Google LLC. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/core/CheckCloudSdkTask.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/core/CheckCloudSdkTask.java
@@ -17,6 +17,7 @@
 
 package com.google.cloud.tools.gradle.appengine.core;
 
+import com.google.cloud.tools.appengine.cloudsdk.AppEngineJavaComponentsNotInstalledException;
 import com.google.cloud.tools.appengine.cloudsdk.CloudSdk;
 import com.google.cloud.tools.appengine.cloudsdk.CloudSdkNotFoundException;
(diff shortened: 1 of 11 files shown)


src/main/java/com/google/cloud/tools/gradle/appengine/core/CheckCloudSdkTask.java
```diff
@@ -0,0 +1,73 @@
...
+  public void setCloudSdkBuilderFactory(CloudSdkBuilderFactory cloudSdkBuilderFactory) {
+    this.cloudSdkBuilderFactory = cloudSdkBuilderFactory;
+  }
+
+  /** Task entrypoint : Download/update/verify Cloud SDK installation. */
+  @TaskAction
+  public void checkCloudSdkAction() {
+    File home = toolsExtension.getCloudSdkHome();
+    if (home == null) {
+      throw new GradleException("SDK home directory must be specified for validation.");
```
TadCordle: This task never runs if `cloudSdkHome` is set in build.gradle, so this should only happen if the user explicitly calls the task without configuring a home directory. Do we want the user to be allowed to run these tasks themselves, or should they only be automatically called?

src/main/java/com/google/cloud/tools/gradle/appengine/core/CheckCloudSdkTask.java
```diff
@@ -0,0 +1,73 @@
...
+
+  public void setToolsExtension(ToolsExtension toolsExtension) {
+    this.toolsExtension = toolsExtension;
+  }
+
+  public void setCloudSdkBuilderFactory(CloudSdkBuilderFactory cloudSdkBuilderFactory) {
+    this.cloudSdkBuilderFactory = cloudSdkBuilderFactory;
+  }
+
+  /** Task entrypoint : Download/update/verify Cloud SDK installation. */
```
coollog: Just `verify`

src/main/java/com/google/cloud/tools/gradle/appengine/core/CheckCloudSdkTask.java
loosebazooka: home == null should never occur should it?

loosebazooka: oh I see what you're saying.

src/main/java/com/google/cloud/tools/gradle/appengine/core/CheckCloudSdkTask.java
```diff
@@ -0,0 +1,73 @@
...
+  @TaskAction
+  public void checkCloudSdkAction() {
+    File home = toolsExtension.getCloudSdkHome();
+    if (home == null) {
+      throw new GradleException("SDK home directory must be specified for validation.");
+    }
+
+    String version = toolsExtension.getCloudSdkVersion();
+    CloudSdk cloudSdk = cloudSdkBuilderFactory.newBuilder().build();
+    if (!version.equals("LATEST") && !version.equals(cloudSdk.getVersion().toString())) {
```
loosebazooka: I don't think LATEST is an actual option for the cloud sdk version. Only the second check is necessary.

TadCordle: but LATEST is an option for the tools extension. If I only do the version.equals() check, it'll fail if the user ever specifies LATEST in their build.gradle.

loosebazooka: Oh I see, I didn't realize a user could specify latest. I don't know if we should include LATEST as an option the user can enter, if they don't specify a version, isn't it implied latest?

TadCordle: Hmm yeah, that is true. I guess I can remove the LATEST option.

loosebazooka:
Can you add a test in "AppEngine<Standard/Flexible>PluginTest"  to make sure this task is included when a cloud SDK home is specified and the download task is not?

PR #219, opened 2018-03-09
Improve cloud SDK installation progress

### LINKED ISSUE #209: Show progress while Cloud SDK is downloading
The `ProgressListener` used in `CloudSdkDownloader` only shows messages at the beginning of installation; it should probably show a progress bar or something while it's being downloaded.

### LINKED ISSUE #214: Pass ProgressListener and ConsoleListener into DownloadCloudSdkTask instead of using anonymous classes

TadCordle:
Fixes #209 and fixes #214. `ProgressListener` and `ConsoleListener` are implemented in separate classes and passed into the task, and installation progress is shown using the `ConsoleListener`.

### DIFF
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
@@ -94,7 +94,7 @@ public class AppEngineCorePluginConfiguration {
                     downloadCloudSdkTask.setCloudSdkBuilderFactory(cloudSdkBuilderFactory);
                     downloadCloudSdkTask.setManagedCloudSdkFactory(managedCloudSdkFactory);
                     downloadCloudSdkTask.setProgressListener(
-                        new DownloadCloudSdkTaskProgressListener(p.getLogger()));
+                        new DownloadCloudSdkTaskProgressListener());
                     downloadCloudSdkTask.setConsoleListener(
                         new DownloadCloudSdkTaskConsoleListener());
 
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
@@ -93,6 +93,10 @@ public class AppEngineCorePluginConfiguration {
                   p -> {
                     downloadCloudSdkTask.setCloudSdkBuilderFactory(cloudSdkBuilderFactory);
                     downloadCloudSdkTask.setManagedCloudSdkFactory(managedCloudSdkFactory);
+                    downloadCloudSdkTask.setProgressListener(
+                        new DownloadCloudSdkTaskProgressListener(p.getLogger()));
+                    downloadCloudSdkTask.setConsoleListener(
+                        new DownloadCloudSdkTaskConsoleListener());
 
                     if (toolsExtension.getCloudSdkHome() == null) {
                       p.getTasks()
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
```diff
@@ -93,6 +93,10 @@ private void createDownloadCloudSdkTask() {
                   p -> {
                     downloadCloudSdkTask.setCloudSdkBuilderFactory(cloudSdkBuilderFactory);
                     downloadCloudSdkTask.setManagedCloudSdkFactory(managedCloudSdkFactory);
+                    downloadCloudSdkTask.setProgressListener(
```
coollog: Could `DownloadCloudSdkTask` make its own `ProgressListener` and `ConsoleListener`?

src/main/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTaskConsoleListener.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package com.google.cloud.tools.gradle.appengine.core;
+
+import com.google.cloud.tools.managedcloudsdk.ConsoleListener;
+
+public class DownloadCloudSdkTaskConsoleListener implements ConsoleListener {
+  @Override
+  public void console(String rawString) {
+    System.out.print(rawString);
```
coollog: Check out Gradle's logging thing: https://docs.gradle.org/current/userguide/logging.html

src/main/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTaskProgressListener.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ */
+
+package com.google.cloud.tools.gradle.appengine.core;
+
+import com.google.cloud.tools.managedcloudsdk.ProgressListener;
+
+class DownloadCloudSdkTaskProgressListener implements ProgressListener {
+  @Override
+  public void start(String message, long totalWork) {}
```
coollog: Will these be filled to not be no-op?

TadCordle: I'm not sure if it's necessary, given how much information the ConsoleListener already gives.

src/main/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTaskConsoleListener.java
TadCordle: Printing to stdout automatically logs at the QUIET level it seems, and I didn't see anything in the gradle api that would let me log without automatically adding a newline (newlines are included in the `rawstring` passed in, so it looks really bad using either `println` or `logger.lifecycle()`).

src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
TadCordle: Hmm, yeah I guess they could. I originally set them externally in case they needed to be mocked in a test, but I can't see much reason to test them the way they are right now. If we end up needing to customize the progress output more, then it may end up being better this way for testing. For now I can simplify it.

src/main/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTaskConsoleListener.java
loosebazooka: Yeah I think system.out.print mimics a console best.

TadCordle: We talked about QUIET maybe not being a good level for it since the `ConsoleListener` output is pretty noisy, so maybe we could buffer the raw string until we hit a newline and then use `logger.lifecycle()`.

I'm going to merge this and just submit another issue for that, though, if you think it's necessary.

PR #223, opened 2018-03-12
Construct ManagedCloudSdk at configuration time

### LINKED ISSUE #216: Avoid modifying CloudSdkBuilderFactory properties within task execution
Home directory is updated after `downloadCloudSdk` task execution. Instead, it should probably be set in the configuration phase. This means the ManagedCloudSdk can be instantiated outside of the task as well, so the ManagedCloudSdkFactory instance doesn't need to be passed around.

### LINKED ISSUE #224: Clean up manual downloadCloudSdk and checkCloudSdk execution
We should either figure out how to prevent the user from running these tasks themselves to prevent them from running with conflicting version/home parameters, or we should more clearly define what these tasks do with all combinations of parameters (e.g. when the user runs downloadCloudSdk with a home directory configured, do we download into the specified directory, run the task as if the user hadn't specified a directory, abort task execution, etc?).

TadCordle:
Simplifies DownloadCloudSdkTask a bit and prevents extension parameters from being modified at execution time. Fixes #216, fixes #224.

### DIFF
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
@@ -87,7 +87,7 @@ public class AppEngineCorePluginConfiguration {
                         ? managedCloudSdk.getSdkHome().toFile()
                         : toolsExtension.getCloudSdkHome());
           } catch (UnsupportedOsException | BadCloudSdkVersionException ex) {
-            throw new GradleException(ex.getMessage());
+            throw new GradleException("Configuring... ", ex);
           }
         });
   }
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
@@ -17,6 +17,9 @@
 
 package com.google.cloud.tools.gradle.appengine.core;
 
+import com.google.cloud.tools.managedcloudsdk.BadCloudSdkVersionException;
+import com.google.cloud.tools.managedcloudsdk.ManagedCloudSdk;
+import com.google.cloud.tools.managedcloudsdk.UnsupportedOsException;
 import org.gradle.api.GradleException;
 import org.gradle.api.Project;
 import org.gradle.util.GradleVersion;
@@ -45,7 +48,7 @@ public class AppEngineCorePluginConfiguration {
   private DeployExtension deployExtension;
   private ToolsExtension toolsExtension;
   private CloudSdkBuilderFactory cloudSdkBuilderFactory;
-  private ManagedCloudSdkFactory managedCloudSdkFactory;
+  private ManagedCloudSdk managedCloudSdk;
   private String taskGroup;
 
   /** Configure core tasks for appengine flexible and standard project plugins. */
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/cloud/tools/gradle/appengine/core/CheckCloudSdkTask.java
```diff
@@ -50,6 +51,10 @@ public void checkCloudSdkAction() {
     }
 
     String version = toolsExtension.getCloudSdkVersion();
+    if (Strings.isNullOrEmpty(version)) {
+      throw new GradleException("SDK version must be specified for validation.");
```
TadCordle: Again, adding this until we figure out how to properly prevent running the task manually (should never happen otherwise).

src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
```diff
@@ -75,9 +78,17 @@ public void configureCoreProperties(
...
-          managedCloudSdkFactory = new ManagedCloudSdkFactory(toolsExtension.getCloudSdkVersion());
+          try {
+            managedCloudSdk =
+                new ManagedCloudSdkFactory(toolsExtension.getCloudSdkVersion()).newManagedSdk();
+            cloudSdkBuilderFactory =
+                new CloudSdkBuilderFactory(
+                    toolsExtension.getCloudSdkHome() == null
+                        ? managedCloudSdk.getSdkHome().toFile()
+                        : toolsExtension.getCloudSdkHome());
+          } catch (UnsupportedOsException | BadCloudSdkVersionException ex) {
```
loosebazooka: I wonder if we can just set the managedCloudSdk conditionally here if `getCloudSdkHome == null`

loosebazooka: We potentially don't want to be throwing a BadCloudSdkVersionException if the user isn't using the managed Cloud SDK.

src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
```diff
@@ -93,10 +104,8 @@ private void createDownloadCloudSdkTask() {
 
               project.afterEvaluate(
                   p -> {
-                    downloadCloudSdkTask.setCloudSdkBuilderFactory(cloudSdkBuilderFactory);
-                    downloadCloudSdkTask.setManagedCloudSdkFactory(managedCloudSdkFactory);
-
                     if (toolsExtension.getCloudSdkHome() == null) {
+                      downloadCloudSdkTask.setManagedCloudSdk(managedCloudSdk);
```
loosebazooka: I'm starting to lean towards just setting this unconditionally (it might be null anyway) from above.

src/main/java/com/google/cloud/tools/gradle/appengine/core/CheckCloudSdkTask.java
```diff
@@ -44,12 +44,11 @@ public void setCloudSdkBuilderFactory(CloudSdkBuilderFactory cloudSdkBuilderFact
   /** Task entrypoint : Verify Cloud SDK installation. */
   @TaskAction
   public void checkCloudSdkAction() {
-    File home = toolsExtension.getCloudSdkHome();
-    if (home == null) {
-      throw new GradleException("SDK home directory must be specified for validation.");
+    if (Strings.isNullOrEmpty(version) || cloudSdkBuilderFactory == null) {
```
loosebazooka: If we make `version` an `@Input` https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/Input.html and the gradle task will fail automatically when this isn't configured.

src/main/java/com/google/cloud/tools/gradle/appengine/core/DownloadCloudSdkTask.java
```diff
@@ -33,32 +31,30 @@
...
   /** Task entrypoint : Download/update Cloud SDK. */
   @TaskAction
   public void downloadCloudSdkAction()
-      throws UnsupportedOsException, BadCloudSdkVersionException, ManagedSdkVerificationException,
-          ManagedSdkVersionMismatchException, InterruptedException, CommandExecutionException,
-          SdkInstallerException, CommandExitException, IOException {
+      throws ManagedSdkVerificationException, ManagedSdkVersionMismatchException,
+          InterruptedException, CommandExecutionException, SdkInstallerException,
+          CommandExitException, IOException {
+    if (managedCloudSdk == null) {
```
loosebazooka: Here too, I think we can delegate to the `@Input` annotation to ensure the property is set.

src/main/java/com/google/cloud/tools/gradle/appengine/core/CheckCloudSdkTask.java
TadCordle: `@Input` is supposed to be used on getters, but these parameters don't have getters (only setters, which are ignored by `@Input`). I don't think it's worth it to add getters just for this.

loosebazooka: Yeah, I mean if we wrote the plugin in groovy you would throw it on the property directly. But for this case, it's fine, using a getter is the standard way to define task inputs.

PR #236, opened 2018-03-29
Add gcloud auth login task and parameter for service account.

loosebazooka:


src/test/java/com/google/cloud/tools/gradle/appengine/core/CheckCloudSdkTaskTest.java
```diff
@@ -88,7 +91,9 @@ public void testCheckCloudSdkAction_versionMismatch() {
   }
 
   @Test
-  public void testCheckCloudSdkAction_sdkInstallationException() {
+  public void testCheckCloudSdkAction_sdkInstallationException()
```
TadCordle: I think this is my mistake, but this should probably be named something more specific since it's only checking for CloudSdkNotFoundException.

src/test/java/com/google/cloud/tools/gradle/appengine/core/CheckCloudSdkTaskTest.java
```diff
@@ -130,7 +141,8 @@ public void testCheckCloudSdkAction_versionFileException() {
   }
 
   @Test
-  public void testCheckCloudSdkAction_appEngineInstallationExceptions() {
+  public void testCheckCloudSdkAction_appEngineInstallationExceptions()
```
TadCordle: Again, I probably could have named this something more specific.

src/test/java/com/google/cloud/tools/gradle/appengine/core/CheckCloudSdkTaskTest.java
```diff
@@ -143,6 +155,8 @@ public void testCheckCloudSdkAction_appEngineInstallationExceptions() {
     } catch (TaskExecutionException ex) {
       Assert.assertEquals(
           ex.getCause().getClass(), AppEngineJavaComponentsNotInstalledException.class);
+    } catch (CloudSdkNotFoundException e) {
```
TadCordle: Should add this to the method signature instead of catching; I think we want the test to fail in this case since we're testing for a different exception. I think this one in particular is already tested in `testCheckCloudSdkAction_sdkInstallationException`.

src/test/java/com/google/cloud/tools/gradle/appengine/core/CheckCloudSdkTaskTest.java
loosebazooka: I think my auto-fix messed up here too.

src/test/java/com/google/cloud/tools/gradle/appengine/core/CheckCloudSdkTaskTest.java
loosebazooka: It looks like we're testing a little too deep here (we probably only need to check that we're calling the checks we want), their behavior is tested by the appengine-plugins-core tests.

Issue #240, opened 2018-04-04
Fix appengineDeployAll task for flexible projects


TadCordle:
Mirror of https://github.com/GoogleCloudPlatform/app-maven-plugin/issues/273

PR #242, opened 2018-04-04
Add appengineDeployAll task for standard projects

### LINKED ISSUE #239: Add appengineDeployAll task for standard project
Mirror of https://github.com/GoogleCloudPlatform/app-maven-plugin/issues/273

TadCordle:
Fixes #239. Still needs some fixes for flexible projects (#240).

### DIFF
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployAllTask.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployAllTask.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2018 Google LLC. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployAllTask.java
```diff
@@ -0,0 +1,62 @@
...
+  public void setCloudSdkBuilderFactory(CloudSdkBuilderFactory cloudSdkBuilderFactory) {
+    this.cloudSdkBuilderFactory = cloudSdkBuilderFactory;
+  }
+
+  /** Task Entrypoint : Deploys the app and all of its config files. */
+  @TaskAction
+  public void deployAllAction() throws AppEngineException {
+    if (!deployConfig.getDeployables().isEmpty()) {
+      getLogger().warn("appengineDeployAll: Ignoring configured deployables.");
+      deployConfig.getDeployables().clear();
```
loosebazooka: I was wondering if we should just implement a copyOf inside deployConfig, so that we don't change state that shouldn't be changed. It seems like it would be safer.

```
public static DeployConfig copyOf(DeployConfig config)
```

loosebazooka: or probably as a copyconstructor

```
public DeployConfig(DeployConfig config)
```

PR #244, opened 2018-04-09
Fix appengineDeployAll for flexible projects

### LINKED ISSUE #240: Fix appengineDeployAll task for flexible projects

TadCordle:
Fixes #240

### DIFF
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
@@ -293,6 +293,7 @@ public class AppEngineCorePluginConfiguration {
 
               project.afterEvaluate(
                   project -> {
+                    // deployConfig is set in AppEngineStandardPlugin and AppEngineFlexiblePlugin
                     deployAllTask.setCloudSdkBuilderFactory(cloudSdkBuilderFactory);
                   });
             });
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
@@ -293,7 +293,6 @@ public class AppEngineCorePluginConfiguration {
 
               project.afterEvaluate(
                   project -> {
-                    deployAllTask.setDeployConfig(new DeployExtension(deployExtension));
                     deployAllTask.setCloudSdkBuilderFactory(cloudSdkBuilderFactory);
                   });
             });
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployAllTask.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployAllTask.java
@@ -48,17 +48,19 @@ public class DeployAllTask extends DefaultTask {
   public void deployAllAction() throws AppEngineException {
     List<File> deployables = deployConfig.getDeployables();
 
-    // Clear out deployables that aren't app.yaml (which is added when task is created)
+    // Clear out configured deployables
     if (!deployables.isEmpty()) {
       deployables.clear();
     }
 
+    // Look for app.yaml
     File appYaml = new File(stageDirectory, "app.yaml");
     if (!appYaml.exists()) {
       throw new GradleException("Failed to deploy all: app.yaml not found.");
     }
     deployables.add(appYaml);
 
+    // Look for configuration yamls
     String[] validYamls = {"cron.yaml", "dispatch.yaml", "dos.yaml", "index.yaml", "queue.yaml"};
     for (String yamlName : validYamls) {
       File yaml = deployConfig.getAppEngineDirectory().toPath().resolve(yamlName).toFile();
@@ -68,6 +70,7 @@ public class DeployAllTask extends DefaultTask {
       }
     }
 
+    // Deploy
     CloudSdk sdk = cloudSdkBuilderFactory.newBuilder(getLogger()).build();
     AppEngineDeployment deploy = cloudSdkBuilderFactory.newAppEngineDeployment(sdk);
     deploy.deploy(deployConfig);
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployAllTask.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployAllTask.java
@@ -45,7 +45,7 @@ public class DeployAllTask extends DefaultTask {
   /** Task Entrypoint : Deploys the app and all of its config files. */
   @TaskAction
   public void deployAllAction() throws AppEngineException {
-    DeployExtension deployCopy = deployConfig.copyOf();
+    DeployExtension deployCopy = new DeployExtension(deployConfig);
     deployCopy.getDeployables().clear();
 
     // Look for app.yaml
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployAllTask.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployAllTask.java
@@ -58,6 +58,7 @@ public class DeployAllTask extends DefaultTask {
     if (!appYaml.exists()) {
       throw new GradleException("Failed to deploy all: app.yaml not found.");
     }
+    getLogger().info("appengineDeployAll: Preparing to deploy app.yaml");
     deployables.add(appYaml);
 
     // Look for configuration yamls
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployAllTask.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployAllTask.java
@@ -45,32 +45,33 @@ public class DeployAllTask extends DefaultTask {
   /** Task Entrypoint : Deploys the app and all of its config files. */
   @TaskAction
   public void deployAllAction() throws AppEngineException {
-    deployConfig.getDeployables().clear();
+    DeployExtension deployCopy = deployConfig.copyOf();
+    deployCopy.getDeployables().clear();
 
     // Look for app.yaml
     File appYaml = stageDirectory.toPath().resolve("app.yaml").toFile();
     if (!appYaml.exists()) {
       throw new GradleException("Failed to deploy all: app.yaml not found.");
     }
-    addDeployable(appYaml);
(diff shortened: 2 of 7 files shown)


src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployAllTask.java
```diff
@@ -36,17 +39,29 @@ public void setCloudSdkBuilderFactory(CloudSdkBuilderFactory cloudSdkBuilderFact
...
+    this.stageDirectory = stageDirectory;
+  }
+
   /** Task Entrypoint : Deploys the app and all of its config files. */
   @TaskAction
   public void deployAllAction() throws AppEngineException {
-    if (!deployConfig.getDeployables().isEmpty()) {
-      getLogger().warn("appengineDeployAll: Ignoring configured deployables.");
-      deployConfig.getDeployables().clear();
+    List<File> deployables = deployConfig.getDeployables();
```
coollog: Would it make sense to instead just call `deployConfig.setDeployables` at the bottom?

src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployAllTask.java
```diff
@@ -36,17 +39,29 @@ public void setCloudSdkBuilderFactory(CloudSdkBuilderFactory cloudSdkBuilderFact
...
+    deployables.add(appYaml);
 
-    String[] validYamls = {
-      "app.yaml", "cron.yaml", "dispatch.yaml", "dos.yaml", "index.yaml", "queue.yaml"
-    };
+    // Look for configuration yamls
+    String[] validYamls = {"cron.yaml", "dispatch.yaml", "dos.yaml", "index.yaml", "queue.yaml"};
     for (String yamlName : validYamls) {
       File yaml = deployConfig.getAppEngineDirectory().toPath().resolve(yamlName).toFile();
       if (yaml.exists()) {
```
coollog: Should there be a "Preparing to deploy app.yaml" message as well?

src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployAllTask.java
TadCordle: Looks like my assertions fail when I do this. I think it's because deployConfig is a mock.

coollog: Ah okay. There's the [`ArgumentCaptor`](https://static.javadoc.io/org.mockito/mockito-core/2.6.9/org/mockito/ArgumentCaptor.html) thing you could use to capture the argument passed into a mocked method and assert things on that. Otherwise, I'd say probably just change L69 to use `deployables` as well just to keep things consistent.

src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployAllTask.java
```diff
@@ -36,17 +39,30 @@ public void setCloudSdkBuilderFactory(CloudSdkBuilderFactory cloudSdkBuilderFact
...
+    if (!deployables.isEmpty()) {
+      deployables.clear();
+    }
+
+    // Look for app.yaml
+    File appYaml = new File(stageDirectory, "app.yaml");
+    if (!appYaml.exists()) {
+      throw new GradleException("Failed to deploy all: app.yaml not found.");
     }
+    getLogger().info("appengineDeployAll: Preparing to deploy app.yaml");
```
coollog: This could be combined with the same log message below in a separate method like you did in the `app-maven-plugin`.

An alternative way to do this would be to structure the flow as:

1. Start list of empty yaml files.
2. Add `app.yaml` file or give exception.
3. Add other yaml files.
4. Check, log, and add files to `deployables`.

src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployAllTask.java
TadCordle: Oh wait, I was looking at the wrong thing. Yeah I think this should work.

src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
```diff
@@ -293,7 +293,6 @@ private void createDeployAllTask() {
 
               project.afterEvaluate(
                   project -> {
-                    deployAllTask.setDeployConfig(new DeployExtension(deployExtension));
```
loosebazooka: can you throw a comment in here saying that it's configured in the individual plugins?

src/test/java/com/google/cloud/tools/gradle/appengine/core/DeployAllTaskTest.java
```diff
@@ -56,31 +54,30 @@
 
   private DeployAllTask deployAllTask;
 
+  private File stageDir;
```
loosebazooka: Do we have a test to ensure the main config isn't changed by this task?

TadCordle: I think that'll need to go somewhere else. It does change what's passed in, we just pass in a copy of the config when the task is created, so I can't really check that in the unit tests.

coollog:
Closing as we found that [`JAVA_TOOL_OPTIONS` is the environment variable that is used by the JVM by default](https://stackoverflow.com/questions/12577437/how-to-set-a-java-system-property-so-that-it-is-effective-whenever-i-start-jvm-w) to do exactly this.

src/test/java/com/google/cloud/tools/gradle/appengine/core/DeployAllTaskTest.java
coollog: Perhaps a safer pattern would be to pass around an immutable `DeployConfiguration` built from `DeployExtension` rather than rely on `#setDeployConfig` to be called with a copy of the mutable `DeployExtension`, but that might be overkill. A simple fix would probably be just moving the `new DeployExtension` call to inside `#setDeployConfig`.

TadCordle: I think the reason I didn't do that originally was so I could mock the internal `deployConfig` in tests.

TadCordle: Maybe we should go with a clone() method instead of a copy constructor?

coollog: One way you could do it is to capture and assert on the `deployConfig` passed into `deploy.deploy` and not use a mocked `DeployExtension` since it doesn't have any side effects.

loosebazooka: @TadCordle oh right I forgot that we put in the config step. Yeah, the no side effects would be ideal. I didn't think of the capture though, that might be the way we want to go.

src/test/java/com/google/cloud/tools/gradle/appengine/core/DeployAllTaskTest.java
```diff
@@ -115,46 +118,57 @@ public void testDeployAllAction_flexible() throws AppEngineException, IOExceptio
...
-    assertFalse(deployConfig.getDeployables().contains(invalidYaml));
-    verify(deploy).deploy(deployConfig);
+    assertTrue(deployCopy.getDeployables().contains(appYaml));
+    assertTrue(deployCopy.getDeployables().contains(cronYaml));
+    assertTrue(deployCopy.getDeployables().contains(dispatchYaml));
+    assertTrue(deployCopy.getDeployables().contains(dosYaml));
+    assertTrue(deployCopy.getDeployables().contains(indexYaml));
+    assertTrue(deployCopy.getDeployables().contains(queueYaml));
+    assertFalse(deployCopy.getDeployables().contains(invalidYaml));
+    verify(deploy).deploy(deployCopy);
```
loosebazooka: I think I might have said conflicting things earlier, but copy constructors are the desired style here,

An option while still using the copy constructor is
```
ArgumentCaptor deployCaptor = ArgumentCaptor.forClass(DeployConfiguration)
verify(deploy).deploy(deployCaptor.capture())
DeployConfiguration captured = deployCaptor.getValue()
assertTrue( //blah blah
```

src/test/java/com/google/cloud/tools/gradle/appengine/core/DeployAllTaskTest.java
```diff
@@ -92,26 +93,85 @@ public void testDeployAllAction() throws AppEngineException, IOException {
...
-    verify(deploy).deploy(deployConfig);
+    verify(deploy).deploy(deployCapture.capture());
+    DeployConfiguration captured = deployCapture.getValue();
+    assertTrue(captured.getDeployables().contains(appYaml));
+    assertTrue(captured.getDeployables().contains(cronYaml));
+    assertTrue(captured.getDeployables().contains(dispatchYaml));
+    assertTrue(captured.getDeployables().contains(dosYaml));
+    assertTrue(captured.getDeployables().contains(indexYaml));
+    assertTrue(captured.getDeployables().contains(queueYaml));
+    assertFalse(captured.getDeployables().contains(invalidYaml));
```
loosebazooka: Do we have the side effect check? `originalDeployConfig.getDeployables == ['original-deployable.yaml']`

loosebazooka: never mind I found it.

src/test/java/com/google/cloud/tools/gradle/appengine/core/DeployAllTaskTest.java
```diff
@@ -92,26 +93,85 @@ public void testDeployAllAction() throws AppEngineException, IOException {
...
+
+    verify(deploy).deploy(deployCapture.capture());
+    DeployConfiguration captured = deployCapture.getValue();
+    assertTrue(captured.getDeployables().contains(appYaml));
+    assertFalse(captured.getDeployables().contains(validInDifferentDirYaml));
+  }
+
+  @Test
+  public void testDeployAllAction_configNotModified() throws AppEngineException, IOException {
+    deployConfig.setAppEngineDirectory(stageDir);
```
loosebazooka: I think this test might cover more bases if it put something in the deployables for the original config.

PR #247, opened 2018-04-11
Skip other deploy tasks when deployAll is also run

### LINKED ISSUE #241: Prevent other deploy tasks from running if deployAll is run

TadCordle:
Fixes #241

### DIFF
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployCronTask.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployCronTask.java
@@ -42,6 +42,16 @@ public class DeployCronTask extends DefaultTask {
   /** Task Entrypoint : deploy cron.yaml. */
   @TaskAction
   public void deployAction() throws AppEngineException {
+    if (getProject()
+        .getGradle()
+        .getTaskGraph()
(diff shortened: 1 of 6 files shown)


coollog:
I am wondering what the case would be to have both individual deploy tasks and deployAll in the same execution graph. Is it safe to not do the other deploys? I'm not sure but what if the execution was something like:

1. deployCron
2. Some task that checks some state after cron is deployed, and could fail
3. deployAll to deploy all the configuration since that check passed

TadCordle:
Good point, but I feel like in that case it'd be better to just configure deployables and use a normal deploy after doing the deployCron+check rather than using deployAll. That way the cron config wouldn't be deployed twice. Are there any cases where deploying the same config twice is necessary?

loosebazooka:
I wonder if the best option is to just the let the user decide, there's no real harm in them deploying it twice.

TadCordle:
Is there any sort of GCP quota people have to worry about if they accidentally deploy too many times too fast? If not I guess this is just to avoid a possible minor annoyance. If you don't think it's necessary I can close the PR.

loosebazooka:
Yeah, I don't think we need to worry about a deploy quota.

PR #251, opened 2018-04-16
Use appengine-web.xml as backup config instead of gcloud global state

### LINKED ISSUE #248: Read application/version from build.gradle + appengine-web.xml instead of build.gradle + gcloud global state
Should verify application property and pass the verified value through to the deploy step.

1. If "deploy.read.appengine.web.xml" system property is set, make sure project id is only set in appengine-web.xml and not build config, otherwise error.
2. If "deploy.read.appengine.web.xml" is not set, then make sure project id is set in build config, otherwise error.

TadCordle:
Fixes #248

### DIFF
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployExtension.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployExtension.java
@@ -19,7 +19,7 @@ package com.google.cloud.tools.gradle.appengine.core;
 
 import com.google.cloud.tools.appengine.api.deploy.DeployConfiguration;
 import com.google.cloud.tools.appengine.api.deploy.DeployProjectConfigurationConfiguration;
-import com.google.cloud.tools.gradle.appengine.standard.ProjectResolver;
+import com.google.cloud.tools.gradle.appengine.standard.PropertyResolver;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.List;
@@ -41,7 +41,7 @@ public class DeployExtension
   private Boolean stopPreviousVersion;
   private String version;
   private File appEngineDirectory;
-  private ProjectResolver projectResolver;
+  private PropertyResolver propertyResolver;
 
   public DeployExtension(Project gradleProject) {
     this.gradleProject = gradleProject;
@@ -60,7 +60,7 @@ public class DeployExtension
     this.stopPreviousVersion = deployExtension.stopPreviousVersion;
     this.version = deployExtension.version;
     this.appEngineDirectory = deployExtension.appEngineDirectory;
-    this.projectResolver = deployExtension.projectResolver;
+    this.propertyResolver = deployExtension.propertyResolver;
   }
 
   @Override
@@ -92,7 +92,7 @@ public class DeployExtension
 
   @Override
   public String getProject() {
-    return (projectResolver == null ? project : projectResolver.getProject(project));
+    return (propertyResolver == null ? project : propertyResolver.getProject(project));
   }
 
   public void setProject(String project) {
@@ -128,7 +128,7 @@ public class DeployExtension
 
   @Override
   public String getVersion() {
-    return (projectResolver == null ? version : projectResolver.getVersion(version));
+    return (propertyResolver == null ? version : propertyResolver.getVersion(version));
   }
 
   public void setVersion(String version) {
@@ -144,7 +144,7 @@ public class DeployExtension
     return appEngineDirectory;
   }
 
-  public void setProjectResolver(ProjectResolver projectResolver) {
-    this.projectResolver = projectResolver;
+  public void setPropertyResolver(PropertyResolver propertyResolver) {
+    this.propertyResolver = propertyResolver;
   }
 }
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployExtension.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployExtension.java
@@ -128,7 +128,7 @@ public class DeployExtension
 
   @Override
   public String getVersion() {
-    return version;
+    return (projectResolver == null ? version : projectResolver.getVersion(version));
   }
 
   public void setVersion(String version) {
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployExtension.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployExtension.java
@@ -17,12 +17,18 @@
 
 package com.google.cloud.tools.gradle.appengine.core;
 
+import com.google.cloud.tools.appengine.AppEngineDescriptor;
+import com.google.cloud.tools.appengine.api.AppEngineException;
 import com.google.cloud.tools.appengine.api.deploy.DeployConfiguration;
 import com.google.cloud.tools.appengine.api.deploy.DeployProjectConfigurationConfiguration;
 import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
+import org.gradle.api.GradleException;
 import org.gradle.api.Project;
+import org.xml.sax.SAXException;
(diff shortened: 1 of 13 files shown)


src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployExtension.java
```diff
@@ -58,6 +65,65 @@ public DeployExtension(DeployExtension deployExtension) {
...
+    }
+
+    // Check system property
+    boolean readAppEngineWebXml =
+        System.getProperty("deploy.read.appengine.web.xml") != null
+            && System.getProperty("deploy.read.appengine.web.xml").equals("true");
+    if (readAppEngineWebXml) {
+      // Use properties from appengine-web.xml if not also set in build.gradle
+      if (project != null && xmlProject != null || version != null && xmlVersion != null) {
+        throw new GradleException(
```
TadCordle: I wonder if this should happen even if the xml properties are empty, i.e. if the system property is set and there's something in the build.gradle, error out, regardless of whether or not there are values to overwrite it with from the xml.

src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployExtension.java
```diff
@@ -40,6 +46,7 @@
   private Boolean stopPreviousVersion;
   private String version;
   private File appEngineDirectory;
+  private File appEngineWebXml;
```
loosebazooka: I missed this whole thing somehow.

src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployExtension.java
```diff
@@ -58,6 +65,65 @@ public DeployExtension(DeployExtension deployExtension) {
...
+      return deployExtensionCopy;
+    }
+
+    AppEngineDescriptor appengineWebXmlDoc =
+        AppEngineDescriptor.parse(new FileInputStream(appEngineWebXml));
+    String xmlProject = appengineWebXmlDoc.getProjectId();
+    String xmlVersion = appengineWebXmlDoc.getProjectVersion();
+
+    // Verify that project and version are set somewhere
+    if (project == null && xmlProject == null || version == null && xmlVersion == null) {
```
loosebazooka: Now that I think about it, I think version can be autogenerated by the cloud sdk.

loosebazooka: we should find out how the IDEs handle this

TadCordle: Yeah, it looks like it can. So I guess we should do the same validation to avoid conflicting build config/appengine-web.xml, but skip the error if it's not configured.

src/main/java/com/google/cloud/tools/gradle/appengine/standard/ProjectResolver.java
```diff
@@ -0,0 +1,85 @@
...
+      // Verify that project and version are set somewhere
+      if (userDefinedProject == null && xmlProject == null) {
+        throw new GradleException(
+            "appengine-plugin does not use gcloud global project state. Please configure the "
+                + "application ID and version in your build.gradle or appengine-web.xml.");
+      }
+
+      // Check system property
+      boolean readAppEngineWebXml =
+          System.getProperty("deploy.read.appengine.web.xml") != null
```
loosebazooka: I think you can use Boolean.getBoolean() here
https://docs.oracle.com/javase/7/docs/api/java/lang/Boolean.html#getBoolean(java.lang.String)

src/main/java/com/google/cloud/tools/gradle/appengine/standard/ProjectResolver.java
```diff
@@ -0,0 +1,85 @@
...
+            return xmlProject;
+          }
+        }
+      } else {
+        // Make sure properties are set in build.gradle
+        if (userDefinedProject == null) {
+          throw new GradleException(
+              "appengine-plugin does not use gcloud global project state. If you would like to "
+                  + "use the state from appengine-web.xml, please set the system property "
+                  + "deploy.read.appengine.web.xml");
```
loosebazooka: `=true`? since we're checking for that

src/main/java/com/google/cloud/tools/gradle/appengine/standard/ProjectResolver.java
```diff
@@ -0,0 +1,85 @@
...
+                + "application ID and version in your build.gradle or appengine-web.xml.");
+      }
+
+      // Check system property
+      boolean readAppEngineWebXml =
+          System.getProperty("deploy.read.appengine.web.xml") != null
+              && System.getProperty("deploy.read.appengine.web.xml").equals("true");
+      if (readAppEngineWebXml) {
+        // Use properties from appengine-web.xml if not also set in build.gradle
+        if (userDefinedProject != null && xmlProject != null) {
```
loosebazooka: Can this be condensed to?
```
if (xmlProject != null) {
  // error if the project override is specified.
  if (userDefindProject != null) {
    // error
  }
  return xmlProject;
}
```

src/main/java/com/google/cloud/tools/gradle/appengine/standard/AppEngineStandardPlugin.java
```diff
@@ -101,14 +102,27 @@ private void configureExtensions() {
             deploy.setAppEngineDirectory(
                 new File(stageExtension.getStagingDirectory(), "WEB-INF/appengine-generated"));
           }
-          deploy.setProjectResolver(
-              new ProjectResolver(
-                  stageExtension
-                      .getStagingDirectory()
+
+          File appengineWebXml =
```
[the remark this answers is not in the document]
TadCordle: I was basing the logic of of `AppEnginePlugin#isAppEngineStandard`, it looks like there's a fallback.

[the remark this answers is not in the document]
loosebazooka: ah yes, good that you brought that up, it's because it's possible at that point the "war" plugin hasn't resolved yet. For instance a project buildfile like this
```
apply plugin "com.google.cloud.tools.appengine"
apply plugin "war"
```
but I think at the point you're resolving it, it shouldn't be a problem, since it's after all plugins have been applied.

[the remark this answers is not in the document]
TadCordle: Ah ok, makes sense.

loosebazooka:
Oops, just want to make sure this is enabled for all the config tasks too deployIndexes... or whatever?

TadCordle:
Hmmm I think I only did deploy and deployAll.

loosebazooka:
also I think you might need to make sure showConfiguration still works?

src/main/java/com/google/cloud/tools/gradle/appengine/standard/PropertyResolver.java
```diff
@@ -0,0 +1,96 @@
...
+    if (readAppEngineWebXml && buildProperty != null) {
+      // Should be reading from appengine-web.xml, but property is configured in build.gradle
+      throw new GradleException(
+          "Cannot override appengine.deploy config with appengine-web.xml. Either remove "
+              + "the project/version properties from your build.gradle, or clear the "
+              + "deploy.read.appengine.web.xml system property to read from build.gradle.");
+    } else if (!readAppEngineWebXml && buildProperty == null && xmlProperty != null) {
+      // Should be reading from build file, but it's only configured in appengine-web.xml
+      throw new GradleException(
+          "appengine-plugin does not use gcloud global project state. If you would like to "
```
loosebazooka: gah, I think this message needs to be slightly different for "version"?

TadCordle:
>also I think you might need to make sure showConfiguration still works?

Created issue #253

PR #252, opened 2018-04-18
Fix error message for appengine-web.xml version

TadCordle:


src/main/java/com/google/cloud/tools/gradle/appengine/standard/PropertyResolver.java
```diff
@@ -86,8 +86,8 @@ private String validatedProperty(String buildProperty, String xmlProperty) {
     } else if (!readAppEngineWebXml && buildProperty == null && xmlProperty != null) {
       // Should be reading from build file, but it's only configured in appengine-web.xml
       throw new GradleException(
-          "appengine-plugin does not use gcloud global project state. If you would like to "
-              + "use the state from appengine-web.xml, please set the system property "
+          "appengine-plugin does not use gcloud global state for project/version. If you would "
```
loosebazooka: I think we want to allow version to passthrough to gcloud?

PR #254, opened 2018-04-19
Fix showConfiguration for project/version

### LINKED ISSUE #253: appengineShowConfiguration shows null for application/version if they're configured in appengine-web.xml

TadCordle:
Fixes #253; instead of reading values directly from fields within extensions, showConfiguration now tries to invoke getters for each field if they exist.


src/main/java/com/google/cloud/tools/gradle/appengine/core/ShowConfigurationTask.java
```diff
@@ -106,12 +110,23 @@ private static String getFieldData(Field root, Object instance, int depth)
...
-        .append(" = ")
-        .append(root.get(instance))
-        .append("\n");
+        .append(" = ");
+    try {
+      Method getter = instance.getClass().getMethod(getGetterFromFieldName(root.getName()));
+      result.append(getter.invoke(instance));
+    } catch (NoSuchMethodException ex) {
+      result.append(root.get(instance));
+    } catch (InvocationTargetException ex) {
```
loosebazooka: how does InvocationTargetException happen?

TadCordle: If there's an exception thrown mid-invocation, it gets wrapped in an InvocationTargetException. In our case it can happen if someone runs showConfiguration w/ misconfigured project or version in appengine-web.xml.

src/main/java/com/google/cloud/tools/gradle/appengine/core/ShowConfigurationTask.java
```diff
@@ -69,7 +72,8 @@ static String getExtensionData(String extensionName, Object extensionInstance, i
         continue;
       }
       // This is just a helper for the extensions, don't show it
-      if (field.getType().equals(org.gradle.api.Project.class)) {
+      if (field.getType().equals(org.gradle.api.Project.class)
```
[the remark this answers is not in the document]
loosebazooka: I think when I wrote this, it was to make it super explicit what the type was that we were skipping.

src/main/java/com/google/cloud/tools/gradle/appengine/core/ShowConfigurationTask.java
loosebazooka: can you add a test to trigger this somehow?

TadCordle: Sure

src/main/java/com/google/cloud/tools/gradle/appengine/core/ShowConfigurationTask.java
```diff
@@ -63,16 +64,18 @@ static String getExtensionData(String extensionName, Object extensionInstance, i
...
+    for (Method method : extensionInstance.getClass().getSuperclass().getDeclaredMethods()) {
       // ignore synthetic fields (stuff added by compiler or instrumenter)
-      if (field.isSynthetic()) {
+      if (method.isSynthetic()) {
         continue;
       }
-      // This is just a helper for the extensions, don't show it
-      if (field.getType().equals(org.gradle.api.Project.class)) {
-        continue;
+      if (method.getName().startsWith("get")
```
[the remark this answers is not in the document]
TadCordle: Does this find getters that do more than just return the corresponding field?

Also, does Gradle map config properties to the fields or to the getters/setters? I feel like just checking how the method is named would be enough, but I'm not sure how Gradle does it (e.g. if it recognizes a method named `isValid` as a getter for a `valid` property or something).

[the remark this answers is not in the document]
TadCordle: @loosebazooka @coollog Is it worth changing this?

PR #258, opened 2018-04-23
Show 2.0 release candidate warning when plugin is applied

### LINKED ISSUE #257: Add a 2.0-release candidate warning when plugin is loaded
I think we want to do something like this, to make sure people don't upgrade blindly.
```
WARNING: You are a using a release candiate ${version}. Behavior of this plugin has change since 1.x.x, please see release notes at : github.com/GoogleCloudPlatform/app-gradle-plugin. Missing a feature? Can't get it to work?, please file a bug at <link to issues>
```

TadCordle:
Fixes #257


src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
```diff
@@ -58,6 +58,19 @@ public void configureCoreProperties(
       Project project,
       AppEngineCoreExtensionProperties appEngineCoreExtensionProperties,
       String taskGroup) {
+    project
+        .getLogger()
+        .warn(
+            "WARNING: You are a using a release candidate for "
```
loosebazooka: I think we might need to work this first sentence another way? Does it render as:
```
You are using a release candidate for 2.0.0-rc1. Behavior....
```

TadCordle: Yeah, I guess that makes it sound like a release candidate for a release candidate.

PR #262, opened 2018-05-01
Remove deployables parameter

### LINKED ISSUE #261: Remove deployables as configuration option

TadCordle:
Fixes #261


src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployExtension.java
```diff
@@ -77,7 +77,9 @@ public void setBucket(String bucket) {
     return deployables;
   }
 
-  public void setDeployables(Object deployables) {
+  // The name of this method is no longer "setDeployables" so the user can't configure it in
+  // their build file.
+  public void updateDeployables(Object deployables) {
```
coollog: Since this will just be an internal method, the parameter type can probably be a list of files.

src/main/java/com/google/cloud/tools/gradle/appengine/flexible/AppEngineFlexiblePlugin.java
```diff
@@ -86,9 +86,8 @@ private void configureExtensions() {
 
           // obtain deploy extension set defaults
           DeployExtension deploy = appengineExtension.getDeploy();
-          if (deploy.getDeployables() == null) {
-            deploy.setDeployables(new File(stageExtension.getStagingDirectory(), "app.yaml"));
-          }
+          deploy.updateDeployables(new File(stageExtension.getStagingDirectory(), "app.yaml"));
```
coollog: Since this is no longer user-configurable, maybe it makes sense to have it just be on the `DeployAllTask` as a setter.

TadCordle: If I did this, I'd still need to set it in the deploy extension somehow. It extends DeployConfiguration, which means I need a `getDeployables()`.

loosebazooka: Yeah, its kind of a weird consequence of how I wrote it initially. DeployExtension doesn't seem to map nicely anymore.

One option is not even have the setter, and have some kind of weird copy constructor that generates a DeployExtension with deployables being a immutable member of DeployExtension.

```
private final List<File> deployables;

...

DeployExtension(DeployExtension deploy, List<File> deployables) {
  // copy stuff as done in the other copy constructor
   this.deployables = ImmutableList.of(deployables)
}
```

remove the setter (or updater), and call this constructor in the task directly (Deploy, DeployAll)?

loosebazooka: and then you'd probably, as Q suggested, pass in the necessary info to the task directly.

src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployAllTask.java
```diff
@@ -45,8 +46,7 @@ public void setStageDirectory(File stageDirectory) {
   /** Task Entrypoint : Deploys the app and all of its config files. */
   @TaskAction
   public void deployAllAction() throws AppEngineException {
-    DeployExtension deployCopy = new DeployExtension(deployConfig);
-    deployCopy.getDeployables().clear();
+    DeployExtension deployCopy = new DeployExtension(deployConfig, new ArrayList<>());
```
loosebazooka: I think this might be a little bit of a strange pattern.

Would it be possible to just populate a local variable list with the files we know about and then call the constructor with that fully populated list?

src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployExtension.java
```diff
@@ -31,9 +31,9 @@
 
   // named gradleProject to disambiguate with deploy parameter "project"
   private final Project gradleProject;
+  private final List<File> deployables;
```
loosebazooka: Could this be an immutable list? http://google.github.io/guava/releases/23.0/api/docs/com/google/common/collect/ImmutableList.html

src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployAllTask.java
TadCordle: It needs the staging directory to resolve the path, so I don't think so. Unless I'm misunderstanding.

src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployExtension.java
TadCordle: I don't think so, the list is cleared/repopulated in deployAll.

loosebazooka: yeah, maybe deployAll could just give the copy constructor a fully populated list.

src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployAllTask.java
loosebazooka: oh, I mean you could do this copy further down in the deployAllAction method. First just create a local list of deployables, and then pass that one to new DeployExtension(deployConfig, deplyables)

TadCordle: Oh I see what you mean.

src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployExtension.java
```diff
@@ -61,6 +60,7 @@ public DeployExtension(DeployExtension deployExtension) {
     this.version = deployExtension.version;
     this.appEngineDirectory = deployExtension.appEngineDirectory;
     this.propertyResolver = deployExtension.propertyResolver;
+    this.deployables = ImmutableList.<File>builder().addAll(deployables).build();
```
loosebazooka: I think you can use `ImmutableList.copyOf` here.

PR #264, opened 2018-05-08
Allow config to be read from everywhere.

loosebazooka:
Introduces some keyworks for 'project' that allow the user
to configure what they want as their source of truth during
deployment

Allow project/version config to be read from
1. build.gradle (this is default)
2. appengine-web.xml : project = 'APPENGINE_CONFIG'
3. gcloud global state: project = 'GCLOUD_CONFIG'

APPENGINE_CONFIG/GCLOUD_CONFIG are not valid google cloud project
names and are reconfigured by the plugin.

Flexible projects do not allow the user to configure APPENGINE_CONFIG


loosebazooka:
@TadCordle looks like we want to allow users to use gcloud global state if they configure it. Here's a version of this stuff that enables it. I initially tried to do it using system properties, but I think this might work out nicer. I probably should have figured this out before you wrote all that code :\.

src/main/java/com/google/cloud/tools/gradle/appengine/flexible/FlexibleDeployTargetResolver.java
```diff
@@ -0,0 +1,66 @@
...
+
+public class FlexibleDeployTargetResolver implements DeployTargetResolver {
+
+  @Override
+  public String getProject(String configString) {
+    if (configString == null
+        || configString.trim().isEmpty()
+        || configString.equals(APPENGINE_CONFIG)) {
+      throw new GradleException(
+          "project must be defined or configured to read from system state\n"
```
TadCordle: Should probably capitalize the start of each sentence here (same for all the other places you throw an exception).

loosebazooka:
@TadCordle yeah we want the same experience in maven. I'll file the bug.

PR #273, opened 2018-06-26
Make it clear the user configured version is wrong

loosebazooka:
Gradle error handling inside an afterEvaluate block produces a confusing result. So make the error message better.


src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
```diff
@@ -95,23 +95,29 @@ public void configureCoreProperties(
...
-          } catch (UnsupportedOsException | BadCloudSdkVersionException ex) {
-            throw new GradleException("Configuring... ", ex);
+          } catch (UnsupportedOsException ex) {
+            throw new RuntimeException(ex.getMessage(), ex);
+          } catch (BadCloudSdkVersionException ex) {
+            throw new RuntimeException(
+                "Failed to auto-configure Cloud Sdk at version = '"
+                    + toolsExtension.getCloudSdkVersion()
+                    + "': "
+                    + ex.getMessage());
```
TadCordle: Should also pass `ex` as the cause parameter, right?

loosebazooka: oh yeah.

PR #280, opened 2018-08-08
Unify build into single gradle file

loosebazooka:
Remove external scripts pulling from build.gradle. This standardizes this with other projects we own.


build.gradle
```diff
@@ -55,6 +44,10 @@ dependencies {
   testCompile 'org.mockito:mockito-core:2.7.21'
 }
 
+task wrapper(type: Wrapper) {
+  gradleVersion = '3.4.1'
```
coollog: Maybe this as well? (to `4.9`)

loosebazooka: hrmm... I think that might trigger some things I'm not ready to clean up :\. Lemme see what happens

PR #288, opened 2018-08-31
Fix source context integration test

loosebazooka:


src/integTest/resources/projects/sourcecontext-project/build.gradle
```diff
@@ -22,3 +22,9 @@ plugins {
   id 'com.google.cloud.tools.source-context'
 }
 
+appengine {
+  deploy {
+    projectId = "project"
```
coollog: nit: should we change the string to `projectId` as well?

loosebazooka: Yeah we could, it's ignored. I guess I might as well.

PR #321, opened 2019-01-08
Resolve GCLOUD_CONFIG/APPENGINE_CONFIG late

loosebazooka:
If a user doesn't have the cloud sdk installed yet, but uses
GCLOUD_CONFIG, it results in a cloud sdk not found error.

While this seems like a strange case, it's safest for us
to only even attempt anything after we can guarantee the cloud
sdk is setup.

### DIFF
--- a/src/main/java/com/google/cloud/tools/gradle/appengine/appyaml/AppYamlDeployTargetResolver.java
+++ b/src/main/java/com/google/cloud/tools/gradle/appengine/appyaml/AppYamlDeployTargetResolver.java
@@ -20,11 +20,12 @@ package com.google.cloud.tools.gradle.appengine.appyaml;
 import static com.google.cloud.tools.gradle.appengine.core.ConfigReader.APPENGINE_CONFIG;
 import static com.google.cloud.tools.gradle.appengine.core.ConfigReader.GCLOUD_CONFIG;
 
-import com.google.cloud.tools.appengine.operations.Gcloud;
+import com.google.cloud.tools.gradle.appengine.core.CloudSdkOperations;
 import com.google.cloud.tools.gradle.appengine.core.ConfigReader;
+import com.google.cloud.tools.gradle.appengine.core.DeployTargetResolver;
 import org.gradle.api.GradleException;
 
-public class AppYamlDeployTargetResolver {
+public class AppYamlDeployTargetResolver implements DeployTargetResolver {
 
   static final String PROJECT_ERROR =
       "Deployment projectId must be defined or configured to read from system state\n"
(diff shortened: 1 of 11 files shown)


src/main/java/com/google/cloud/tools/gradle/appengine/core/DeployExtension.java
```diff
@@ -45,21 +47,28 @@ public DeployExtension(Project gradleProject) {
     this.gradleProject = gradleProject;
   }
 
+  public void setDeployTargetResolver(DeployTargetResolver deployTargetResolver) {
+    this.deployTargetResolver = deployTargetResolver;
+  }
+
   DeployConfiguration toDeployConfiguration(List<Path> deployables) {
+    String processedProjectId = deployTargetResolver.getProject(projectId);
+    String processedVersion = deployTargetResolver.getProject(version);
```
chanseokoh: `getVersion()`?

src/main/java/com/google/cloud/tools/gradle/appengine/standard/AppEngineStandardPlugin.java
```diff
@@ -123,11 +119,9 @@ private void configureExtensions() {
...
                   .toFile();
 
-          // configure the deploy extensions's project/version parameters
-          StandardDeployTargetResolver resolver =
-              new StandardDeployTargetResolver(appengineWebXml, cloudSdkOperations.getGcloud());
-          deploy.setProjectId(resolver.getProject(deploy.getProjectId()));
-          deploy.setVersion(resolver.getVersion(deploy.getVersion()));
+          // configure the supplier for project/version parameters
+          deploy.setDeployTargetResolver(
+              new StandardDeployTargetResolver(appengineWebXml, cloudSdkOperations));
```
chanseokoh: What do you think of saving the `new StandardDeployTargetResolver` instance in a local var, and reusing it below in `runExtension.setDeployTargetResolver()`?

src/test/java/com/google/cloud/tools/gradle/appengine/standard/StandardDeployTargetResolverTest.java
```diff
@@ -49,6 +50,7 @@
   private File appengineWebXml;
 
   @Mock Gcloud gcloud;
+  @Mock CloudSdkOperations cloudSdkOperations;
```
chanseokoh: `private`

src/main/java/com/google/cloud/tools/gradle/appengine/appyaml/AppYamlDeployTargetResolver.java
```diff
@@ -46,24 +47,25 @@
...
    * allowed for app.yaml based projects), show usage. If set to GCLOUD_CONFIG then read from
    * gcloud's global state. If set but not a keyword then just return the set value.
    */
+  @Override
   public String getProject(String configString) {
     if (configString == null
         || configString.trim().isEmpty()
         || configString.equals(APPENGINE_CONFIG)) {
       throw new GradleException(PROJECT_ERROR);
     } else if (configString.equals(GCLOUD_CONFIG)) {
```
coollog: tiny nit: can remove `else`s

loosebazooka: oh yeah, weird that I did that, must be losing my mind.

PR #325, opened 2019-01-29
Remove options of using APPENGINE_CONFIG

loosebazooka:
Users of all types of deployments must use either an actual
value or `GCLOUD_CONFIG` to delegate to gcloud's configured state.


src/main/java/com/google/cloud/tools/gradle/appengine/standard/AppEngineStandardPlugin.java
```diff
@@ -144,7 +138,7 @@ private void configureExtensions() {
           if (Strings.isNullOrEmpty(runExtension.getProjectId())) {
             runExtension.setProjectId(deploy.getProjectId());
           }
-          runExtension.setDeployTargetResolver(standardResolver);
+          runExtension.setDeployTargetResolver(deploy.getDeployTargetResolver());
```
chanseokoh: Will `deploy.getDeployTargetResolver()` return null? A new `deploy` instance is created in this method, and I don't immediately see anyone calling `setDeployTargetResolver()` on `deploy`.

src/test/java/com/google/cloud/tools/gradle/appengine/core/DeployTargetResolverTest.java
```diff
@@ -5,26 +5,23 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *         http://www.apache.org/licenses/LICENSE-2.0
+ *          http://www.apache.org/licenses/LICENSE-2.0
```
TadCordle: Were these two changes (here and L15) intentional?

src/main/java/com/google/cloud/tools/gradle/appengine/standard/AppEngineStandardPlugin.java
loosebazooka: Yeah, that's where the code is a little wonky. It might worth just instantiating a new one for clarification. But, it is created in another area where we create the Deploy Extension: https://github.com/GoogleCloudPlatform/app-gradle-plugin/pull/325/files#diff-61ff2d5be7163076a4c1810bd162f73eR129

I think I might just create a new object here.

src/test/java/com/google/cloud/tools/gradle/appengine/core/DeployTargetResolverTest.java
loosebazooka: No, I don't know what happened, it might be the license file I have in intellij? I'll revert them for now.

PR #326, opened 2019-02-12
Update to plugins core 0.7.4

loosebazooka:


loosebazooka:
NOTE that this includes a change from plugins core that reverts java 11 standard-compat support.

PR #330, opened 2019-02-27
Fix bullet

shiraji:


TadCordle:
I believe these were intended to be footnotes, not bullets.

shiraji:
OK

shiraji:
image

but this seems really hard read as a footnote...

suztomo:
Thank you for feedback. I also see the footnote (2nd one) is broken.

Changing
“property. * setting”  -> “property, while setting”
will make them look better

suztomo:
Fixed in #331.

PR #333, opened 2019-03-15
Clean up code

chanseokoh:
@loosebazooka see if this makes sense. Hopefully no one is using `cloudSdkOperations` (it's `private`) through reflection.


loosebazooka:
Yeah must be artifacts from a rewrite.

PR #338, opened 2019-04-16
Release v2.0.0 rc6

loosebazooka:


chanseokoh:
Shouldn't we Increase the SNAPSHOT version to 2.0.1-SNAPSHOT? Nothing is being changed.

PR #343, opened 2019-05-01
Add support for publishing to gradle plugin portal

loosebazooka:


loosebazooka:
I think we might need to rearrange some logic to get it to work with the plugin portal :\

PR #351, opened 2019-06-14
Remove Kokoro sign and publish scripts

chanseokoh:


chanseokoh:
Also updating Travis to use OpenJDK 8 instead of Oracle JDK. Recently I've seen Travis having trouble installing Oracle JDK in another repo.

PR #364, opened 2019-10-22
Upgrade plugins-core to 0.8.1

loosebazooka:
Adds extended support for thin jar deployments, automatically copying
jars referenced in Class-Path in the source artifact to the staging
directory.


loosebazooka:
looks like we're not going to run into the mavencentral problem when building with gradle.

PR #384, opened 2020-09-09
Add verbosity option

loosebazooka:
Verified manually that gcloud app deploy run from plugin respect verbosity flag
Can really find a good way to test this without reflection. I guess I could do that?


src/main/java/com/google/cloud/tools/gradle/appengine/core/CloudSdkOperations.java
```diff
@@ -42,7 +42,7 @@
    * @param credentialFile optional path to a credential file
```
chanseokoh, line 42: Update the Javadoc?

chanseokoh (approved):

PR #388, opened 2020-09-21
Allow configuration using the plugins block

loosebazooka:
Must ensure `war` is applied before we work on our plugin. We could rewrite the whole plugin to configure on demand and use the necessary late evaluation mechanism but this is sufficient for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 
+## Unreleased
+### Fixed
+* Fixed bug when using `plugins` block in `build.gradle` ([#388](../../pull/388))
+
 ## 2.4.0
 ### Added
 * `appengine.tools.verbosity` option for defining gcloud log verbosity ([#384](../../pull/384))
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -20,10 +20,10 @@ automatically installed by the plugin.
 
 # How to use
 
-The plugin JAR needs to be defined in your build script. It is directly available on Maven Central. Alternatively, you can download it from GitHub and deploy it to your local repository. The following code snippet shows an example on how to retrieve it from Maven Central:
+The plugin needs to be defined in your `build.gradle`. It is directly available on Maven Central. Alternatively, you can download it from GitHub and deploy it to your local repository. The following code snippet shows an example on how to retrieve it from Maven Central:
 
 ### Using plugins block
-Since the appengine-gradle-plugin is not published to the gradle plugin portal, you must refrence it from the Central Maven repository.  Add the maven central resolution to your `settings.gradle`
+The `appengine-gradle-plugin` is not published to the gradle plugin portal, you must refrence it from the Central Maven repository.  Add the maven central resolution to your `settings.gradle`
```
chanseokoh, line 26: I think "since ..." is correct unless you terminate the statement by a period like "... to the gradle plugin portal."? And "since" sounds natural to me.

chanseokoh (approved):

README.md
loosebazooka: Yeah I was supposed to add a `so` after the the comma, but I'll just revert to `Since`

loosebazooka:
Forgot readme updates should be after a release. Going to split this into 2 PRs.

PR #389, opened 2020-09-21
Post 2.4.1 docs

loosebazooka:


USER_GUIDE.md
```diff
@@ -1,36 +1,25 @@
...
-apply plugin: "com.google.cloud.tools.appengine-appyaml"
-```
 
 When you use the `com.google.cloud.tools.appengine` plugin it will automatically determine
 your environment based on the presence of an `appengine-web.xml`
 in `src/main/webapp/WEB-INF/`. It will enable _`appengine-appenginewebxml`_ if present, _`appengine-appyaml`_ otherwise.
 
 The [Cloud SDK](https://cloud.google.com/sdk) is required for this plugin to
-function. Download and install it before running any tasks.
+function. Versions greater than `2.0.0` will automatically download and install it for you. You may also chose to download and install it before running any tasks.
```
chanseokoh, line 15:
```suggestion
function. Versions greater than `2.0.0` will automatically download and install it for you. You may also choose to download and install it before running any tasks.
```

README.md
```diff
@@ -20,27 +20,51 @@ automatically installed by the plugin.
...
-apply plugin: 'com.google.cloud.tools.appengine'
+pluginManagement {
+  repositories {
+    gradlePluginPortal()
+    mavenCentral()
+    // add mavenLocal() if you are using a locally built version of the plugin
+  }
+  resolutionStrategy {
+    eachPlugin {
+      if (requested.id.id.startsWith('com.google.cloud.tools.appengine')) {
```
chanseokoh, line 36: Maybe `requested.id.id` --> `requested.id`?

And `resolutionStrategy` and `useModule` are alien to me. Why do we need them? Are they general required whenever someone wants to apply a plugin from the Maven Central?

src/main/java/com/google/cloud/tools/gradle/appengine/standard/AppEngineStandardPlugin.java
```diff
@@ -61,6 +61,7 @@
   @Override
   public void apply(Project project) {
     this.project = project;
+    project.getPluginManager().apply(WarPlugin.class);
```
chanseokoh, line 64: Just in case, this change is already in #388.

loosebazooka: Oh right, I built this off that branch, I'll rebase it away,

README.md
loosebazooka: Right, so `requested.id.id` is correct.

Basically there's no mapping from the plugin "com.google.cloud.tools.appengine" to the artifact. So here we create that mapping.

[`ResolutionStrategy.eachPlugin`](https://github.com/gradle/gradle/blob/674c1da6b12978247560a3223670cc1835482f1c/subprojects/plugin-use/src/main/java/org/gradle/plugin/management/internal/DefaultPluginResolutionStrategy.java#L46)

uses the type [`PluginResolveDetails`](https://github.com/gradle/gradle/blob/1293df631c1a20647a0489175d7aa63d60d65e30/subprojects/core-api/src/main/java/org/gradle/plugin/management/PluginResolveDetails.java#L32)

where we get a [`PluginRequest`] (https://github.com/gradle/gradle/blob/1293df631c1a20647a0489175d7aa63d60d65e30/subprojects/core-api/src/main/java/org/gradle/plugin/management/PluginRequest.java#L36)

And the first `id` is the `id` of the request
in the style
```
plugins {
  id `the-id-in-question` version `...`
}
```
This `id` is actually a [`PluginId`](https://github.com/gradle/gradle/blob/1293df631c1a20647a0489175d7aa63d60d65e30/subprojects/core-api/src/main/java/org/gradle/plugin/use/PluginId.java)

So we obtain the `id` from this `id`, and hence the nested `requested.id.id`

USER_GUIDE.md
chanseokoh: Still ”chose”.

chanseokoh (approved):

README.md
loosebazooka: > Are they general required whenever someone wants to apply a plugin from the Maven Central?

Also to answer this no. We can publish a plugin to maven central in the same way as one would to gradle plugin repo, but it's a two step process.

1. Publish the actual jar with code: https://plugins.gradle.org/m2/gradle/plugin/com/google/cloud/tools/jib-gradle-plugin/2.5.0/
2. Publish the "plugin jar" following some gradle convention with a link to the jar in (1): 
https://plugins.gradle.org/m2/com/google/cloud/tools/jib/com.google.cloud.tools.jib.gradle.plugin/2.5.0/

So I think gradle can do the same for other repos like maven central -- as long as we follow the convention?

chanseokoh: Now I get it. The Gradle plugin model requiring a separate "plugin jar" seems a bit weird to me, but I guess there must be reason.

I'm curious why the App Engine plugin is not published to the Gradle Plugin Portal but to Maven Central (like Jib)?

chanseokoh (approved):

PR #394, opened 2021-02-05
Avoid creating and configuring tasks when unnecessary

### LINKED ISSUE #393: app-gradle-plugin breaks tasks in multi project build?
I want to use the [frontend-gradle-plugin](https://github.com/siouan/frontend-gradle-plugin) with this plugin.

If I start with the [multi-project-war-application example](https://github.com/siouan/frontend-gradle-plugin/tree/master/examples/multi-projects-war-application) from the fronend-gradle-plugin everything builds fine.

When I add the app-gradle-plugin using either the method in the documentation the build the fails. For example if I add it through a build script block:

```
buildscript {
  repositories {
    mavenCentral()
  }

  dependencies {
    classpath 'com.google.cloud.tools:appengine-gradle-plugin:2.4.1'
  }
}

apply plugin: 'com.google.cloud.tools.appengine'
```

Gradle will fail with:
```
A problem occurred configuring project ':backend'.
> Could not create task ':backend:processFrontendResources'.
   > Task with name 'assembleFrontend' not found in project ':frontend'.

* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Exception is:
org.gradle.api.ProjectConfigurationException: A problem occurred configuring project ':backend'.
	at org.gradle.configuration.project.LifecycleProjectEvaluator.wrapException(LifecycleProjectEvaluator.java:75) 
...
```

Note that in the example project the build.gradle has the following part:
```
tasks.register('processFrontendResources', Copy) {
    // Directory containing the artifacts in the frontend project
    def frontendBuildDir = file("${project(':frontend').buildDir}/www")
    // Directory where the frontend artifacts must be copied to be packaged alltogether with the backend by the 'war'
    // plugin.
    def frontendResourcesDir = file("${project.buildDir}/resources/main/public")

    group 'Frontend'
    description 'Process frontend resources'
    dependsOn project(':frontend').tasks.named('assembleFrontend')

    from frontendBuildDir
    into frontendResourcesDir
}
```

After adding the app-gradle-plugin the dependsOn line fails. Through debugging I've found that project(':frontend').tasks becomes empty after the app-gradle-plugin is added.

What is the app-gradle-plugin doing that is changing tasks in the project? Is there a different way I can look up the assembleFrontend task that will work with the app-gradle-plugin or is the plugin fundamenatally incompatible with multi project gradle builds?

chanseokoh:
Fixes GoogleCloudPlatform/appengine-plugins#1005.

Based on https://docs.gradle.org/current/userguide/task_configuration_avoidance.html


chanseokoh:
However, this still doesn't work because `project.afterEvaluate()` is forbidden in the new configuration avoidance API. From https://docs.gradle.org/current/userguide/task_configuration_avoidance.html, it says

> 8. Some APIs may be disallowed if you try to access them from the new API’s configuration blocks. For example, `Project.afterEvaluate()` cannot be called when configuring a task registered with the new API.

So when actually applied to a project, it fails:
```
* What went wrong:
A problem occurred configuring project ':backend'.
> Failed to notify project evaluation listener.
   > Could not create task ':backend:downloadCloudSdk'.
      > Project#afterEvaluate(Action) on project ':backend' cannot be executed in the current context.
   > Could not create task ':backend:appengineDeployAll'.
      > Project#afterEvaluate(Action) on project ':backend' cannot be executed in the current context.
```

chanseokoh:
Fixes the problem, and the integration tests are passing. However, I don't have strong confidence that it doesn't break anything. Previously, even after the integration tests passing, I encountered an error on a sample project. For now, I'll leave this in the draft state.

src/main/java/com/google/cloud/tools/gradle/appengine/appyaml/AppEngineAppYamlPlugin.java
```diff
@@ -89,10 +91,10 @@ private void configureExtensions() {
           // we can only set the default location of "archive" after project evaluation (callback)
           if (stageExtension.getArtifact() == null) {
             if (project.getPlugins().hasPlugin(WarPlugin.class)) {
-              War war = (War) project.getProperties().get("war");
+              War war = project.getTasks().withType(War.class).getByName("war");
```
chanseokoh, line 94: For some reason that I don't know, the previous code was returning null.

Also I think the new code may still unnecessarily create the `war` task.

src/main/java/com/google/cloud/tools/gradle/appengine/appyaml/AppEngineAppYamlPlugin.java
```diff
@@ -124,27 +126,28 @@ private void configureExtensions() {
...
+            STAGE_TASK_NAME,
+            StageAppYamlTask.class,
+            stageTask -> {
+              stageTask.setGroup(APP_ENGINE_APP_YAML_TASK_GROUP);
+              stageTask.setDescription("Stage an App Engine app.yaml based project for deployment");
+              stageTask.dependsOn(BasePlugin.ASSEMBLE_TASK_NAME);
+            });
+    project.afterEvaluate(
+        project -> {
+          if (stage.isPresent()) {
```
chanseokoh, line 141: Actually, this check doesn't seem to work as I expected. It seems to always return true, whether the task has been actually created or not. :(

src/main/java/com/google/cloud/tools/gradle/appengine/appyaml/AppEngineAppYamlPlugin.java
```diff
@@ -89,10 +91,10 @@ private void configureExtensions() {
           // we can only set the default location of "archive" after project evaluation (callback)
           if (stageExtension.getArtifact() == null) {
             if (project.getPlugins().hasPlugin(WarPlugin.class)) {
-              War war = (War) project.getProperties().get("war");
+              War war = tasks.withType(War.class).getByName("war");
```
chanseokoh, line 94: For some reason that I don't know, the previous code was returning null.

Also I think the new code may still unnecessarily create the war task.

chanseokoh:
I think this should do it.

src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
```diff
@@ -121,217 +123,164 @@ private void configureFactories() {
...
+                taskClass,
+                task -> {
+                  task.setGroup(taskGroup);
+                  task.setDescription(taskDescription);
+                });
+
+    if (injectDeployExtension) {
+      project.afterEvaluate(
+          project -> taskProvider.configure(task -> task.setDeployExtension(deployExtension)));
+    }
```
vlsi, line 271: What is the purpose of `afterEvaluate` here?
It looks like `task.setDeployExtension(deployExtension)` can be called during task registration.

src/main/java/com/google/cloud/tools/gradle/appengine/appyaml/AppEngineAppYamlPlugin.java
vlsi: Right you are. It does instantiate create task.
The way to avoid task creation is `stageExtension.setArtifact(tasks.named<War>(WarPlugin.WAR_TASK_NAME).map { it.getArchiveFile() })`

src/main/java/com/google/cloud/tools/gradle/appengine/appyaml/AppEngineAppYamlPlugin.java
```diff
@@ -89,10 +91,10 @@ private void configureExtensions() {
...
               stageExtension.setArtifact(war.getArchivePath());
             } else if (project.getPlugins().hasPlugin(JavaPlugin.class)) {
-              Jar jar = (Jar) project.getProperties().get("jar");
+              Jar jar = tasks.withType(Jar.class).getByName("jar");
```
vlsi, line 97: Same here: it is better to use `tasks.named(...)`

src/main/java/com/google/cloud/tools/gradle/appengine/appyaml/AppEngineAppYamlPlugin.java
```diff
@@ -107,44 +109,43 @@ private void configureExtensions() {
...
-              (DeployTask)
-                  project.getTasks().getByName(AppEngineCorePluginConfiguration.DEPLOY_TASK_NAME);
-          deployTask.setDeployConfig(deploy);
-          deployTask.setAppYaml(stageExtension.getStagingDirectory().toPath().resolve("app.yaml"));
+          tasks
+              .withType(DeployAllTask.class)
+              .configureEach(
+                  task -> {
+                    task.setStageDirectory(stageExtension.getStagingDirectory());
+                    task.setDeployExtension(deploy);
```
vlsi, line 117: If you use `Property<...>` for `stageExtension.getStagingDirectory`, then you could move this task configuration outside of `afterEvaluate`.

src/main/java/com/google/cloud/tools/gradle/appengine/appyaml/AppEngineAppYamlPlugin.java
```diff
@@ -107,44 +109,43 @@ private void configureExtensions() {
...
+        tasks.register(
+            STAGE_TASK_NAME,
+            StageAppYamlTask.class,
+            task -> {
+              task.setGroup(APP_ENGINE_APP_YAML_TASK_GROUP);
+              task.setDescription("Stage an App Engine app.yaml based project for deployment");
+              task.dependsOn(BasePlugin.ASSEMBLE_TASK_NAME);
+            });
+    project.afterEvaluate(
+        project -> stage.configure(task -> task.setStagingConfig(stageExtension)));
```
vlsi, line 142: Is `afterEvaluate` needed here?

src/main/java/com/google/cloud/tools/gradle/appengine/appyaml/AppEngineAppYamlPlugin.java
chanseokoh: But the code below needs to do `Jar.getArchivePath()`, which I believe will require instantiating the task anyway, right?

chanseokoh: Oh sorry, nvm. Didn't see the other comment.

src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
chanseokoh: I have the same question. (I have never worked on this repo, and I don't have a lot of Gradle knowledge.)

What I can see is that `deployExtension` is instantiated in the following way in `Plugin.apply()`.

First, a parent extension is created with `.create()`
```
    appengineExtension =
        project.getExtensions().create("appengine", AppEngineAppYamlExtension.class);
```
and then the sub-extension (`this` below is `appengineExtension` created above)
```
    deploy =
        ((ExtensionAware) this).getExtensions().create(DEPLOY_EXT, DeployExtension.class, project);
```
This `deploy` is what is called `deployExtension` in this class (if I am following the code flow correctly).

So the question is, do we still need to use `afterEvaluate()` when calling `setDeployExtension(deployExtension)`? I guess not, because it's just passing and setting a Java reference. But since I have never worked on this repo, I'm concerned there may be something that I'm not seeing. WDYT?

src/main/java/com/google/cloud/tools/gradle/appengine/appyaml/AppEngineAppYamlPlugin.java
chanseokoh: Sounds like a great suggestion, but unfortunately, I don't really follow what you are talking about. (I don't have much experience with Gradle.) And perhaps this kind of refactoring can be done incrementally in a separate PR. Maybe you could file an issue with an explanation of a potential solution?

src/main/java/com/google/cloud/tools/gradle/appengine/core/AppEngineCorePluginConfiguration.java
vlsi: >I guess not, because it's just passing and setting a Java reference

Indeed. If the reference is not changed, then it can be passed without `afterEvaluate`.

chanseokoh: I removed more `afterEvaluate` blocks in 1ca1a13 and 29190de.

src/main/java/com/google/cloud/tools/gradle/appengine/appyaml/AppEngineAppYamlPlugin.java
chanseokoh: Removed `getByName()` in another place in bae6dc2. I assume `getProject().zipTree(object)` will work fine when given a `TaskProvider`. (Previously, it was getting `File` directly from `getArchivePath()`.) Does this look good?

InLaw:
What can be done to continue here, who needs to do the review?

src/main/java/com/google/cloud/tools/gradle/appengine/appyaml/AppEngineAppYamlPlugin.java
```diff
@@ -89,11 +90,11 @@ private void configureExtensions() {
...
             if (project.getPlugins().hasPlugin(WarPlugin.class)) {
-              War war = (War) project.getProperties().get("war");
-              stageExtension.setArtifact(war.getArchivePath());
+              stageExtension.setArtifact(
+                  tasks.withType(War.class).named("war").map(war -> war.getArchivePath()));
             } else if (project.getPlugins().hasPlugin(JavaPlugin.class)) {
-              Jar jar = (Jar) project.getProperties().get("jar");
-              stageExtension.setArtifact(jar.getArchivePath());
+              stageExtension.setArtifact(
+                  tasks.withType(Jar.class).named("jar").map(jar -> jar.getArchivePath()));
```
TWiStErRob, line 97: no need to create a filtered task collection:
```suggestion
                  tasks.named("jar", Jar.class).map(jar -> jar.getArchivePath()));
```

PR #396, opened 2021-02-08
Migrate to GitHub Actions for unit tests and set up code coverage

chanseokoh:
The GitHub Actions environment has `gcloud` pre-installed.


mpeddada1 (approved):

chanseokoh:
Looks like Jacoco is not compatible with Java 11. Needs more investigation.

mpeddada1:
Hm it looks like Jacoco started officially supporting java 11 from version [0.8.3](https://github.com/jacoco/jacoco/releases). Is there a possibility that we're using an older version here?

chanseokoh:
Looks like it. The reason is unclear, but explicitly setting the version fixes it.

```groovy
jacoco {
  toolVersion = '0.8.6'
}
```

PR #404, opened 2021-07-28
Add note about custom webAppDirName breaking autodetection

### LINKED ISSUE #403: Plugin detection doesn't support overridden web directory
It seems it looks for `src/main/webapp`, however `war` plugins supports overriding the location with:
```
webAppDirName = 'web'
```

It confuses the plugin and `appengine-web.xml` is not detected. By default, the yml version is selected, which doesn't have the `appengineRun` taks that is referenced in the documentation.

Workaround: specify the xml plugin explicitly:
apply plugin: 'com.google.cloud.tools.appengine-appenginewebxml'

elefeint:
Fixes #403.

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -10,8 +10,8 @@ Apply the plugin following the instructions in [README.md](README.md#how-to-use)
 When you use the `com.google.cloud.tools.appengine` plugin it will automatically determine
 your environment based on the presence of an `appengine-web.xml`
 in `src/main/webapp/WEB-INF/`. It will enable _`appengine-appenginewebxml`_ if present, _`appengine-appyaml`_ otherwise.
-Note: autodetection will not work if you use a custom `webAppDirName` property.
-In this case, manually list the environment-specific plugin in your configuration.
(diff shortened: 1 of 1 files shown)


USER_GUIDE.md
```diff
@@ -10,6 +10,8 @@ Apply the plugin following the instructions in [README.md](README.md#how-to-use)
 When you use the `com.google.cloud.tools.appengine` plugin it will automatically determine
 your environment based on the presence of an `appengine-web.xml`
 in `src/main/webapp/WEB-INF/`. It will enable _`appengine-appenginewebxml`_ if present, _`appengine-appyaml`_ otherwise.
+Note: autodetection will not work if you use a custom `webAppDirName` property.
+In this case, manually list the environment-specific plugin in your configuration.
```
chanseokoh, line 14:
```suggestion
Note: autodetection will not work if you set a custom [`webAppDirName` property](https://docs.gradle.org/current/userguide/war_plugin.html#sec:war_convention_properties) of the `war` plugin.
In this case, manually apply the right environment-specific plugin by specifying the corresponding plugin identifier as above.
```

chanseokoh (approved):

PR #421, opened 2022-07-25
Update version in README.md

martinbonnin:
The page at https://cloud.google.com/appengine/docs/standard/java-gen2/using-gradle will also need updating, not sure how to do this?


mpeddada1 (approved):
Thank you for keeping our docs up-to-date @martinbonnin!

PR #422, opened 2022-07-25
Remove dead links

martinbonnin:
These 2 links were 404 now. Feel free to replace with the new links.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,10 +75,6 @@ options.
 
 # Reference Documentation
 
-App Engine Standard Environment:
-* [Using Gradle and the App Engine Plugin (standard environment)](https://cloud.google.com/appengine/docs/java/tools/gradle)
-* [App Engine Gradle Plugin Tasks and Parameters (standard environment)](https://cloud.google.com/appengine/docs/java/tools/gradle-reference)
-
 App Engine Flexible Environment:
 * [Using Gradle and the App Engine Plugin (flexible environment)](https://cloud.google.com/appengine/docs/flexible/java/using-gradle)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -75,10 +75,6 @@ options.
 
 # Reference Documentation
 
-App Engine Standard Environment:
-* [Using Gradle and the App Engine Plugin (standard environment)](https://cloud.google.com/appengine/docs/java/tools/gradle)
```
chanseokoh, line 79: These should work:
https://cloud.google.com/appengine/docs/standard/java/using-gradle
https://cloud.google.com/appengine/docs/standard/java/gradle-reference

@GoogleCloudPlatform/cloud-java-team-teamsync the links for the Maven plugin are also broken. But this is weird; I remember these used to work recently. The page has the banner

> Notice: Over the next few months, we're reorganizing the App Engine documentation site to make it easier to find content and better align with the rest of Google Cloud products. The same content will be available, but the navigation will now match the rest of the Cloud products.

and the actual link redirects to `.../docs/legacy/standard/...`. I wonder if the breakage is related.

chanseokoh: Ah, they also have "-gen2" docs that I think focus on post-Java 8. So we have these links for "using Gradle."

https://cloud.google.com/appengine/docs/standard/java/using-gradle
https://cloud.google.com/appengine/docs/legacy/standard/java/using-gradle
https://cloud.google.com/appengine/docs/standard/java-gen2/using-gradle

mpeddada1: I think you're right, @chanseokoh. Created a PR to fix these links: https://github.com/GoogleCloudPlatform/app-gradle-plugin/pull/424

elefeint:
Thanks! This is superceded by #424 and #473.

PR #423, opened 2022-07-25
Move build scripts to Kotlin

martinbonnin:
This is a follow up from https://github.com/GoogleCloudPlatform/app-gradle-plugin/pull/398

This PR converts build scripts to Kotlin. I tried as much as possible to preserve the existing behaviour even if it's not really idiomatic at the moment. 
This PR also bumps the Gradle version to 6.9. This is the last version that ships with the "maven" plugin and still supports Kotlin DSL.

Follow ups:
* Migrate "maven" to "maven-publish"
* Make the build more idiomatic


build.gradle.kts
```diff
@@ -0,0 +1,219 @@
...
+tasks.register("integTest", Test::class.java) {
+  testClassesDirs = sourceSets.getByName("integTest").output.classesDirs
+  classpath = sourceSets.getByName("integTest").runtimeClasspath
+  outputs.upToDateWhen { false }
+}
+/* TESTING */
+
+
+/* RELEASING */
+tasks.register("sourceJar", Jar::class.java) {
```
vlsi, line 103:
```suggestion
tasks.register<Jar>("sourceJar") {
```

build.gradle.kts
```diff
@@ -0,0 +1,219 @@
...
+              "scm", mapOf(
+                "url" to "https://github.com/GoogleCloudPlatform/app-gradle-plugin",
+                "connection" to "scm:https://github.com/GoogleCloudPlatform/app-gradle-plugin.git",
+                "developerConnection" to "scm:git://github.com/GoogleCloudPlatform/app-gradle-plugin.git"
+              )
+            )
+
+            invokeMethod("licenses", delegateClosureOf<groovy.lang.GroovyObject> {
+              invokeMethod("license", delegateClosureOf<groovy.lang.GroovyObject> {
+                setProperty("name", "The Apache Software License, Version 2.0")
```
vlsi, line 141: Did you mean `withGroovyBuilder` here?
https://docs.gradle.org/current/userguide/kotlin_dsl.html#the_kotlin_dsl_groovy_builder

It would be something like

```kotlin
withGroovyBuilder {
    "licenses" {
        "license" {
            "name"("The Apache Software License, Version 2.0")
            "url"("http://www.apache.org/licenses/LICENSE-2.0.txt")
            ...
        }
    }
}
```

build.gradle.kts
```diff
@@ -0,0 +1,217 @@
...
+          }
+        }.writeTo(outputFile)
+      }
+    }
+  }
+}
+
+
+// for kokoro releases
+tasks.register("prepareRelease", Copy::class.java) {
```
vlsi, line 160:
```suggestion
tasks.register<Sync>("prepareRelease") {
```

Most likely, "sync" task was assumed here, so the outcome does not carry stale files produced by copy tasks from the past executions.

build.gradle.kts
```diff
@@ -0,0 +1,217 @@
...
+/* TESTING */
+
+
+/* RELEASING */
+tasks.register<Jar>("sourceJar") {
+  from(sourceSets.main.get().allJava)
+  classifier = "sources"
+}
+
+tasks.register("javadocJar", Jar::class.java) {
```
vlsi, line 108:
```suggestion
tasks.register<Jar>("javadocJar") {
```

build.gradle.kts
```diff
@@ -0,0 +1,217 @@
...
+    runtimeClasspath += main.get().output
+  }
+}
+
+configurations {
+  named("integTestCompile").get().extendsFrom(testCompileClasspath.get())
+  named("integTestRuntime").get().extendsFrom(testRuntimeClasspath.get())
+}
+
+tasks.register("integTest", Test::class.java) {
```
vlsi, line 94:
```suggestion
tasks.register<Test>("integTest") {
```

`::class.java` is a code smell in `kts`.

martinbonnin: It's technically the same thing though, right?

vlsi: It is, however, `<Test>` is less verbose than `Test::class.java`, and it conveys the intention better.
An alternative option is to use

```
val javadocJar by tasks.registering(Jar::class) {
...
}
```

Then you would be able to replace `tasks.named("javadocJar")` with a reference to `javadocJar` variable.
It is still the same thing, yet referencing the variable makes it easier to follow the code.

martinbonnin: 👍

elefeint:
@jamesward Would you like to help review this PR?

build.gradle.kts
```diff
@@ -0,0 +1,214 @@
...
+
+  testCompile("commons-io:commons-io:2.4")
+  testCompile("junit:junit:4.12")
+  testCompile("org.hamcrest:hamcrest-library:1.3")
+  testCompile("org.mockito:mockito-core:2.23.4")
+}
+
+
+tasks.wrapper {
+  gradleVersion = "6.9"
```
jamesward, line 56: Why the older version?

build.gradle.kts
```diff
@@ -0,0 +1,214 @@
...
+project.afterEvaluate {
+  tasks.register("writePom") {
+    val outputFile = file("$buildDir/pom/${project.name}-${project.version}.pom")
+    outputs.file(outputFile)
+
+    doLast {
+      maven {
+        pom {
+          project {
+            withGroovyBuilder {
```
jamesward, line 123: There is a cleaner syntax for this that would be preferable.  Like:
https://github.com/jamesward/gha-central-test/blob/fbf486aab2b6d60632a8c2b078457be1a05e1a74/build.gradle.kts#L21

vlsi: Would you consider merging #398 then ? https://github.com/GoogleCloudPlatform/app-gradle-plugin/pull/398/files#diff-c0dfa6bc7a8685217f70a860145fbdf416d449eaff052fa28352c5cec1a98c06R252

jamesward: Looks like that is a draft but also it seems like the goal of this is to minimize the changes, then with another PR clean things up / make them more Kotlin-like.  In which case, I'm good with the Groovy style for now.

jamesward: And sorry, I should have read the description in this PR.  :)

vlsi:
@martinbonnin , you might want to add an intermediate commit that renames `build.gradle -> build.gradle.kts` to help Git for displaying the history: https://stackoverflow.com/questions/44310285/converting-file-tracked-with-git-from-java-to-kotlin-in-android-studio/44320452#44320452

build.gradle.kts
martinbonnin: 6.9 is the last Gradle version that still has the (now removed) "maven" plugin. 
This PR is basically taken from https://github.com/GoogleCloudPlatform/app-gradle-plugin/pull/398 and doing less things at a time but I'm happy to switch to https://github.com/GoogleCloudPlatform/app-gradle-plugin/pull/398 directly

build.gradle.kts
martinbonnin: For the record, I'm happy if we close this PR and merge #398 directly. A lot of things have changed in publishing land lately.

martinbonnin:
>  you might want to add an intermediate commit that renames build.gradle -> build.gradle.kts 

Good call 👍 . Also makes it easier to review the diff with the Groovy version from [that commit](https://github.com/GoogleCloudPlatform/app-gradle-plugin/pull/423/commits/e93a839f28635c425379cb20d1425f933354f67f)

build.gradle.kts
elefeint: #398 does too much. Let's take this one step at a time.

martinbonnin:
> Could you rebase this off latest?

It should be good

elefeint (approved):
Thank you!

build.gradle.kts
```diff
@@ -16,177 +20,195 @@
...
-  dependsOn build
-  dependsOn cleanPrepareRelease
+  dependsOn(tasks.build)
 }
 
 release {
-  tagTemplate = 'v$version'
-  git {
-    requireBranch = /^release-v\d+.*$/  //regex
+  tagTemplate = "v$version"
```
vlsi, line 172: This should have been `"v\$version"`

build.gradle.kts
```diff
@@ -16,177 +20,195 @@
...
-    attributes 'Implementation-Title': project.name,
-        'Implementation-Version': version,
-        'Built-By': System.getProperty('user.name'),
-        'Built-Date': new Date(),
-        'Built-JDK': System.getProperty('java.version'),
-        'Built-Gradle': gradle.gradleVersion
+    attributes(
+      mapOf(
+        "Implementation-Title" to project.name,
+        "Implementation-Version" to version,
```
vlsi, line 64: This should have been `project.version` rather than `version` from a jar task (there's a `version` property in `jar` task)

build.gradle.kts
elefeint: In #437

build.gradle.kts
elefeint: I'll add it to #437; thanks!

elefeint:
I verified that releasing process still works and the staged artifacts look good (I've dropped them, so no new release right now). Thanks, everyone!

vlsi:
@elefeint , I see you squashed the commits before merging. It causes broken history since Git can't detect that `build.gradle -> build.gradle.kts` was rename+modifications.

The intermediate commit was vital for proper git history, see my comment in https://github.com/GoogleCloudPlatform/app-gradle-plugin/pull/423#issuecomment-1200123593.

Would you please clarify if "rebase and merge" (with or without merge commit) was an option at all?

PR #437, opened 2022-09-07
Fix tagTemplate

elefeint:
Another follow-up to #423 -- double quotes lead to incorrect resolution of tag template name, leaving "-SNAPSHOT" in the tag title.

Found solution in https://github.com/researchgate/gradle-release/issues/211


mpeddada1 (approved):

martinbonnin (approved):
Thanks for catching this!

Issue #446, opened 2023-01-03
Deprecated Gradle API usage: AbstractArchiveTask.archivePath


TWiStErRob:
<details><summary>The AbstractArchiveTask.archivePath property has been deprecated.</summary>

```
> Configure project :deploy:appengine
The AbstractArchiveTask.archivePath property has been deprecated. This is scheduled to be removed in Gradle 9.0. Please use the archiveFile property instead. See https://docs.gradle.org/8.0-rc-1/dsl/org.gradle.api.tasks.bundling.AbstractArchiveTask.html#org.gradle.api.tasks.bundling.AbstractArchiveTask:archivePath for more details.
	at org.gradle.api.tasks.bundling.AbstractArchiveTask.getArchivePath(AbstractArchiveTask.java:159)
	at org.gradle.api.tasks.bundling.Jar_Decorated.getArchivePath(Unknown Source)
	at com.google.cloud.tools.gradle.appengine.appyaml.AppEngineAppYamlPlugin.lambda$configureExtensions$0(AppEngineAppYamlPlugin.java:96)
	at org.gradle.configuration.internal.DefaultUserCodeApplicationContext$CurrentApplication$1.execute(DefaultUserCodeApplicationContext.java:123)
	at org.gradle.configuration.internal.DefaultListenerBuildOperationDecorator$BuildOperationEmittingAction$1.run(DefaultListenerBuildOperationDecorator.java:171)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$1.execute(DefaultBuildOperationRunner.java:29)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$1.execute(DefaultBuildOperationRunner.java:26)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:66)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:59)
	at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:157)
	at org.gradle.internal.operations.DefaultBuildOperationRunner.execute(DefaultBuildOperationRunner.java:59)
	at org.gradle.internal.operations.DefaultBuildOperationRunner.run(DefaultBuildOperationRunner.java:47)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor.run(DefaultBuildOperationExecutor.java:68)
[...]

PR #448, opened 2023-01-13
Fix deprecation warnings in build

TWiStErRob:
See individual commit messages for details.

The only deprecation warnings that remain are from https://github.com/GoogleCloudPlatform/app-gradle-plugin/pull/398 and #446, so those are not in scope here.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -41,14 +41,14 @@ java {
 group = "com.google.cloud.tools"
 
 dependencies {
-  compile(localGroovy())
-  compile(gradleApi())
-  compile("com.google.cloud.tools:appengine-plugins-core:0.9.9")
+  implementation(localGroovy())
+  implementation(gradleApi())
+  api("com.google.cloud.tools:appengine-plugins-core:0.9.9")
 
-  testCompile("commons-io:commons-io:2.4")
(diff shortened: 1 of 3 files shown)


build.gradle.kts
```diff
@@ -102,13 +108,13 @@ tasks.register<Test>("integTest") {
 /* RELEASING */
 tasks.register<Jar>("sourceJar") {
   from(sourceSets.main.get().allJava)
-  classifier = "sources"
+  archiveClassifier.set("sources")
```
emmileaf, line 111: Dropping a note here on the switch to archiveClassifier for posterity: https://discuss.gradle.org/t/publishing-jars-with-classifiers-after-v5-0/30144/4

emmileaf (approved):
LGTM - Thanks for these fixes @TWiStErRob!

build.gradle.kts
TWiStErRob: Note: https://docs.gradle.org/7.6/dsl/org.gradle.api.tasks.bundling.Jar.html#org.gradle.api.tasks.bundling.Jar:classifier
is deprecated and replacement has `Property<String>` type which you cannot assign ([yet](https://github.com/gradle/gradle/pull/22240), but soon).

PR #449, opened 2023-01-13
Bump testing libraries to their latest possible versions

TWiStErRob:
See individual commits for details, copied here for convenience:

---

- Dependency version bumps:
  - Bump commons-io from 2.4 to 2.11.0
    * Replace usage of deprecated `readFileToString(File)` with `readFileToString(File, Charset)`
  - Bump junit from 4.12 to 4.13
    * Replace usages of deprecated `junit.Assert.assertThat` with `hamcrest.MatcherAssert.assertThat`
  - Bump hamcrest-library from 1.3 to 2.2
  - Bump mockito-core from 2.11.0 to 4.11.0
    * [Mockito 5.0](https://github.com/mockito/mockito/releases/tag/v5.0.0) (released at the same weeks as I raised the PR) requires Java 11 bytecode, so it's not viable yet, because CI runs on [8,11].
  - Bump jacoco from 0.8.6 to 0.8.8
  - Bump checkstyle from 8.18 to 8.37
    * 8.38 requires fixing 12 MissingJavadocType failures on public main source set classes.
    * 9.3 can be used once 8.38 javadocs are fixed, because that's the latest version that supports Java 8. 
    * Done up to here in https://github.com/GoogleCloudPlatform/app-gradle-plugin/pull/452
    * 10.x requires Java 11 in Gradle runtime, while the .github/workflows/unit-tests.yaml is executed on Java 8, Checkstyle 10+ is not viable.
- Have `verifyGoogleFormat` task run before checkstyle tasks (to suggest fixing via `./gradlew googleJavaFormat`)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -192,12 +192,7 @@ googleJavaFormat {
 tasks.check.configure {
   dependsOn(tasks.verifyGoogleJavaFormat)
 }
-tasks.checkstyleMain.configure {
-  // Set up a soft dependency so that verifyGoogleFormat suggests running googleJavaFormat,
-  // before devs start fixing individual checkstyle violations manually.
-  shouldRunAfter(tasks.verifyGoogleJavaFormat)
-}
-tasks.named("checkstyleIntegTest").configure {
+tasks.withType<Checkstyle>().configureEach {
   // Set up a soft dependency so that verifyGoogleFormat suggests running googleJavaFormat,
   // before devs start fixing individual checkstyle violations manually.
   shouldRunAfter(tasks.verifyGoogleJavaFormat)
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -189,8 +189,8 @@ tasks.check.configure {
 // to auto-format run ./gradlew googleJavaFormat
 
 checkstyle {
-  toolVersion = "8.18"
-  // get the google_checks.xml file from the actual tool we"re invoking)
+  toolVersion = "8.37"
+  // Get the google_checks.xml file from the actual tool we're invoking.
   config = resources.text.fromArchiveEntry(configurations.checkstyle.files.first(), "google_checks.xml")
   maxErrors = 0
   maxWarnings = 0
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -45,7 +45,7 @@ dependencies {
   compile(gradleApi())
   compile("com.google.cloud.tools:appengine-plugins-core:0.9.9")
 
-  testCompile("commons-io:commons-io:2.4")
+  testCompile("commons-io:commons-io:2.11.0")
   testCompile("junit:junit:4.12")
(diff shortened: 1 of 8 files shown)


build.gradle.kts
```diff
@@ -186,11 +186,21 @@ googleJavaFormat {
...
+tasks.checkstyleMain.configure {
+  // Set up a soft dependency so that verifyGoogleFormat suggests running googleJavaFormat,
+  // before devs start fixing individual checkstyle violations manually.
+  shouldRunAfter(tasks.verifyGoogleJavaFormat)
+}
+tasks.named("checkstyleIntegTest").configure {
+  // Set up a soft dependency so that verifyGoogleFormat suggests running googleJavaFormat,
+  // before devs start fixing individual checkstyle violations manually.
+  shouldRunAfter(tasks.verifyGoogleJavaFormat)
+}
```
emmileaf, line 198: Nit: if I'm understanding the intention here correctly, perhaps something like this could also work?
```
tasks.withType<Checkstyle> {
  shouldRunAfter(tasks.verifyGoogleJavaFormat)
}
```

emmileaf (commented):
@TWiStErRob Thanks for making these updates! Overall LGTM, apologies for the hassle that some conflicts will need to be resolved here since we've merged #448.

The individual commits were very helpful in tracking the different changes - I’ve updated the PR description with details according to your commit messages for better visibility, please edit or comment if there is anything I am missing or miscapturing here.

build.gradle.kts
TWiStErRob: https://github.com/TWiStErRob/gcp-gradle-plugin/commit/e09b7e0502cc309d3774dbd18c9205c60666b7c3

TWiStErRob: Note: never ever ever ever use `withType<> {}` it's eager Gradle configuration (it'll create all tasks).

`withType<>().configureEach {}` is the lazy new API from [docs](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:old_vs_new_configuration_api_overview):
image

TWiStErRob: I'm planning on opening a PR eventually to make this plugin lazy(er).

TWiStErRob:
> apologies for the hassle that some conflicts will need to be resolved here

No worries, I caused the conflicts, I made them auto-resolvable in IDEA. Done in https://github.com/GoogleCloudPlatform/app-gradle-plugin/pull/449/commits/9c55a6af48e1d3e052b958c21fdae84a0c7d2f07.

build.gradle.kts
emmileaf: > Note: never ever ever ever use `withType<> {}` it's eager Gradle configuration (it'll create all tasks).

Ah, thanks for catching this and replacing with the alternative! TIL :)

emmileaf (approved):
Thank you for the contribution!

